PR #44, opened 2022-05-10
fix cast error and change unit of measure of distance, fix #15

### LINKED ISSUE #15: make_holidays: incompatibility of date type and date range fix
When fixing the distance variables for the beginning and end of the date range
distance[distance>183] = 365 - distance[distance>183]

UFuncTypeError: Cannot cast ufunc 'greater' input 0 from dtype('<m8[ns]') to dtype('<m8') with casting rule 'same_kind'

pietroppeter:
issue explanation:
- distance is initially computed as a timedelta in nanoseconds
- the fix for end of date range `distance[distance > 183] = 365 - distance[distance > 183]` fails in the comparison `distance > 183` and is actually assuming that distance is in days

fix explanation:
- distance is explictly converted to be measure in days

related change:
- removed the conversion of distance to be measured in minutes, it will stay measured in days (I do not think this would affect the rest, but I might be mistaken)

### DIFF
--- a/tsfeatures/calendar/make_holidays.py
+++ b/tsfeatures/calendar/make_holidays.py
@@ -98,13 +98,11 @@ def distance_to_holiday(holiday_dates, dates):
 
     # Compute day distance to holiday
     distance = np.expand_dims(dates_np, axis=1) - np.expand_dims(holiday_dates_np, axis=0)
+    # convert to distance in days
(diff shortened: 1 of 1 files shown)


FedericoGarza:
Hey! Closing this PR since this was related to a previous repo version. Now, the repo contains the python SDK to access TimeGPT. :)

PR #111, opened 2023-09-23
[FIX] Dependencies and local data

FedericoGarza:
This PR fixes CI issues related to directly generating test data. Instead of that, the data is now read from GitHub.


FedericoGarza:
OMG finally

PR #136, opened 2023-10-27
add docs scripts

jmoralez:
Adds the action that builds the mintlify docs and pushes them to the docs branch on every push to main along with the scripts needed to create them.


AzulGarza (approved):
thank you @jmoralez🎉

PR #154, opened 2023-11-08
checkout docs repo instead of using submodule

jmoralez:
Changes the process to build the docs from using a submodule to cloning the repo.


AzulGarza (approved):
😍

PR #225, opened 2024-02-15
set upper bound on pandas

jmoralez:
Sets an upper bound on pandas while https://github.com/ray-project/ray/issues/42842 is fixed.


AzulGarza (approved):
awesome @jmoralez!🎉 thank you

PR #251, opened 2024-03-14
FEAT - information about partitions are now logged

marcopeix:
The number of partitions is logged
We also log which partition is being treated

Screenshot 2024-03-14 at 2 42 49 PM


nixtlats/nixtla_client.py
```diff
@@ -749,12 +749,15 @@ def partition_by_uid(func):
...
         if num_partitions is None or num_partitions == 1:
             return func(self, **kwargs, num_partitions=1)
+        main_logger.info(f"Number of partitions: {num_partitions}")
         df = kwargs.pop("df")
         X_df = kwargs.pop("X_df", None)
         id_col = kwargs["id_col"]
         uids = df["unique_id"].unique()
         results_df = []
+        split_index = 1
         for uids_split in np.array_split(uids, num_partitions):
```
AzulGarza, line 759: just a minor style comment, we could use  `enumerate`:

```python
splits = np.array_split(uids, num_partitions)
for split_index, uids_split in enumerate(splits, start=1):
     ...
```

non-blocking comment.

AzulGarza (approved):

nixtlats/nixtla_client.py
```diff
@@ -749,12 +749,15 @@ def partition_by_uid(func):
...
+            main_logger.info(f"Partition {split_index} of {num_partitions}")
```
AzulGarza, line 760: i just realized, if `len(uids) < num_partitions`, the logger can print incorrect partitions. i think it might be better to use:

```
splits = np.array_split(uids, num_partitions)
for ...:
     main_logger.info(f"Partition {split_index} of {len(splits)}")
```

AzulGarza (changes requested):
awesome @marcopeix! i left a minor comment. thank you! 🙌

nixtlats/nixtla_client.py
```diff
@@ -767,6 +770,7 @@ def wrapper(self, num_partitions, **kwargs):
                 kwargs_uids["X_df"] = X_df_uids
             results_uids = func(self, **kwargs_uids, num_partitions=1)
             results_df.append(results_uids)
+            split_index += 1
```
AzulGarza, line 773: could you please remove this line? it is no longer needed since `enumerate` updates `split_index` in each iteration.  :)

AzulGarza (changes requested):

nixtlats/nixtla_client.py
marcopeix: Good catch! I removed it!

marcopeix:
Is this feature still relevant since partitions are now processed in parallel? The current output looks like this:
Screenshot 2024-04-19 at 10 24 46 AM

mergenthaler:
@marcopeix, closing this for grooming. If you feel it should be kept open, please re-open and correct.

Issue #294, opened 2024-04-16
Feat: Using historical multivariate timeseries in Timegpt


DanKil82:
I am currently checking `TimeGPT` with multivariate timeseries data and comparing it against univariate timeseries data, hoiwever there is no difference in the forecasts.

- The univariate data has 3 cols: date, timeseries_id and the target value
- The multivaraite data has an additional column with numeric values, lets call it `mv_column`
- I know about exogeneous vars, which is supported in TimeGPT, but the `mv_column`is purely historical.

my Code:
```python
# some imports
...

    timegpt_client = TimeGPT(
        api_key=api_key
    )
    timegpt_forecast = timegpt_client.forecast(
        df = df_train,
        h = NixtlaParams.forecast_horizon,
        freq = NixtlaParams.frequency,
        time_col = NixtlaParams.time_column,
        target_col = NixtlaParams.target_column,
        id_col = NixtlaParams.key_column,
    )

```

How can i integrate multivariate features ? Or is this simply not supported ?

elephaint:
Hi,

historical exogenous inputs are not supported in TimeGPT; i.e. TimeGPT only leverages exogenous inputs from `df` if you also supply future values for those exogenous inputs in `X_df`. Otherwise, as you witnessed, the exogenous inputs are ignored. We will add a warning for this behavior.

DanKil82:
Thanks for the feedback. For now i think a warning would be appropriate. 

Is it planned to add this functionality ?, I can just recommend to do so (if possible). I am running a forecasting platform in a large scale industrial company serving  use cases thoughout Demand Planning, Material Planning, Manufacturing PLanning etc. and what i see, that most of my internal customers want to add historical exogeneous data, e.g. price changes, inventory levels etc.

elephaint:
Thanks for the suggestion; Unfortunately I don't have a timeline for that atm. For now, I'd suggest to make a forecast for these exogenous too - this can be rather simple (e.g. by setting them to the last value for the entire forecast horizon, for example for price or price change), or complex by having a separate model / pipeline predicting those values.

Note that the added benefit of this is that you will be explicitly including your assumptions about these variables in the model, whereas the assumptions on these variables will be implicitly modelled if only used as a historical exogenous variable.

PR #295, opened 2024-04-18
[FIX] Raise warning on missing X_df when df has exogenous

elephaint:
Adds a warning that exogenous features in `df` are ignored if `X_df` is not provided. 

Fix for #294 until we support historical exogenous variables in TimeGPT.


nixtlats/nixtla_client.py
```diff
@@ -269,6 +269,11 @@ def transform_inputs(self, df: pd.DataFrame, X_df: pd.DataFrame):
                 X_df = X_df.assign(unique_id="ts_0")
             if X_df.dtypes.ds != "object":
                 X_df["ds"] = X_df["ds"].astype(str)
+        elif set(["unique_id", "ds", "y"]) < set(df.columns):
+            main_logger.warning(
+                "You did not provide X_df. Exogenous features in df are ignored."
+            )
+
```
AzulGarza, line 276: thanks @elephaint! since exogenous variables can be included for some methods, for example in `detect_anomalies` (see [this section of our docs](https://docs.nixtla.io/docs/anomaly_detection#exogenous-variables)), it might be better to include the warning in the [`preprocess_dataframe`](https://github.com/Nixtla/nixtla/blob/41298468045324fee30d510f39e6acd3f223deb7/nixtlats/nixtla_client.py#L430) function which handles this scenario (see for example, that the exogenous variables are used when `X_df is None and self.h is None`, the later case indicating the usage of `detect_anomalies`).  it might be a good idea to include the list of missing exogenous variables and a hint on how to solve the warning. 

also, please add a small test. thank you!🙌

AzulGarza (changes requested):

nixtlats/nixtla_client.py
elephaint: Ok, changed the location and message, and added a test at the bottom of `nixtla_client.ipynb`

nixtla/nixtla_client.py
```diff
@@ -491,6 +492,17 @@ def preprocess_dataframes(
             X_df = df.drop(columns="y")
             x_cols = X_df.drop(columns=["unique_id", "ds"]).columns.to_list()
             X_df = self.preprocess_X_df(X_df)
+
+        if (X_df is None) and (set(y_cols) < set(df.columns)):
+            missing_exogenous = df.drop(columns=y_cols).columns
```
AzulGarza, line 497: it might be better to use `df.columns.drop(y_cols)` since `df.drop` creates a copy of the dataframe.

elephaint: Good point, fixed

AzulGarza (approved):

PR #297, opened 2024-04-23
Handle latex equations in markdown

marcopeix:
Modifying the markdown now handles Latex equations.

The behavior is similar to how codeblocks are handled:

- Find all Latex equations using regex
- Replace them with a placeholder
- Replace the placeholder with the raw equation string
- Do some string cleaning to ensure the equation is displayed properly
- Works for both inline and block equations

Here is an example of an equation after the modify_markdown function is done:

$\mathrm{MAE}(\mathbf{y}_{\tau}, \mathbf{\hat{y}}_{\tau}) = \frac{1}{H} \sum^{t+H}_{\tau=t+1} |y_{\tau} - \hat{y}_{\tau}|$

In the notebook containing these equations, I used small headers instead of bold text for better rendering on the docs.


marcopeix:
It's possible that the added function does not fix the rendering issues. From the image below, it seems that Latex is not supported on docs generated by readme.com. So I used images instead.
Screenshot 2024-04-23 at 2 28 46 PM

AzulGarza (changes requested):
thank you so much @marcopeix! great work. i've published the new docs from this branch:
image

it would be great if we could:

- remove the `equation` text behind each equation.
- fix the problem with the lists.
- fix the space between the paragraph and the next section so the `####` header can be rendered properly. 

ideally, these changes will help also render correctly other sections of the documentation, for example, in the [calendar features tutorial](https://docs.nixtla.io/docs/holidays_and_special_dates):

image

I'll share on slack how you can deploy the docs to readme from a branch.

marcopeix:
@mergenthaler @AzulGarza , I don't if this PR is still useful, but I use .png files to render the equations using codecogs. This way, the background is always white and the equations are readable in both light and dark mode. If it's not necessary, ignore this PR!

Result looks like this: 
Screenshot 2024-04-29 at 4 30 56 PM

mergenthaler:
@marcopeix, can we close this? Or is this still unresolved? If yes, please link to the corresponding PR and close this.

marcopeix:
Not required since this [PR](https://github.com/Nixtla/nixtla/pull/324) was merged. Closing

PR #298, opened 2024-04-23
rename module to nixtla

jmoralez:
* Renames the python module to `nixtla`. 
* Updates the python-publish action to also publish the `nixtlats` package, which will issue a warning on import.
* Removes the `action_files` folder in the setup, since it was being included in the project.
* Adds an action (`test-publish.yml`) that can be triggered manually to release the package to [Test PyPI](https://test.pypi.org/) to test the release process.
* pins the dask version to prevent some tests from failing, will try to fix in a separate PR.


AzulGarza (approved):
awesome @jmoralez! 🥳🎊🫶

PR #300, opened 2024-04-23
call corresponding endpoint in cross_validation

jmoralez:
Calls the cross_validation endpoint in the API when using the cross_validation method.

This required a couple of refactors:
* The logic to determine the date features inside the preprocess_dataframes was moved to add_date_features to be able to call that method directly, because the logic in that method isn't compatible with the structure needed for cross_validation. This also prevented some attributes from being overriden to reduce the complexity and prevent unwanted interactions.
* The restrict_input logic was extracted to a separate method.


nixtla/nixtla_client.py
```diff
@@ -30,18 +30,11 @@
...
-    join,
-    maybe_compute_sort_indices,
-    take_rows,
-    vertical_concat,
-)
 
 from .client import Nixtla
 from .core import ApiError
 from .types.multi_series_anomaly import MultiSeriesAnomaly
+from .types.multi_series_cross_validation import MultiSeriesCrossValidation
```
AzulGarza, line 37: pfff beautiful change

nixtla/nixtla_client.py
```diff
@@ -536,6 +529,24 @@ def validate_input_size(self, Y_df: pd.DataFrame):
...
+            #  to reduce latency when
+            #  a user passes level. since currently the model
+            #  uses conformal prediction, we can change a minimum
+            #  amount of data if the series are too large
+            new_input_size = 3 * self.input_size + max(self.model_horizon, self.h)
+        else:
+            # we only want to forecast
+            new_input_size = self.input_size
+        return new_input_size
+
```
AzulGarza, line 549: nice

AzulGarza (approved):
awesome @jmoralez! lvgtm🥳

PR #308, opened 2024-04-25
regenerate fern code to support pydantic v1 and v2

jmoralez:
* Moves the fern configuration to the repo instead of relying on external pull requests.
* Updates the generated code to be compatible with pydantic v1 and v2.
* Creates a workflow that will verify that the generated code matches the expected.
* Removes the `pydantic<2` restriction in `setup.py`


AzulGarza (approved):
🥳 lgtm @jmoralez! i've opened a pr in our api repo to update the ci/cd to write `timegpt.json` here in future releases.

PR #313, opened 2024-04-25
convert ids to categorical

jmoralez:
Converts the ids to categoricals, extracts their codes and uses that to build the payloads to reduce their sizes. Once we get the response we restore the codes back to their original values.


AzulGarza (approved):
😎 lgtm

PR #314, opened 2024-04-25
feat: restrict input forecast

AzulGarza:
this pr restricts the input to the minimum amount of data required (if available) to speed up api calls using the `forecast` method without requesting prediction intervals, finetunning, or the usage of exogenous variables. 

a test ensuring accuracy and time was performed using the `action_files/models_performance` module. here are the results: 

image

this PR reduces the required `action_files/models_performance` time by 5 seconds while keeping the same accuracy. we expect this PR to accelerate requests, especially when dealing with very large time series.


nixtla/nixtla_client.py
```diff
@@ -559,17 +559,23 @@ def forecast(
...
-            and self.level is not None
-            and not add_history
-        )
+        restrict_input = self.finetune_steps == 0 and X_df is None and not add_history
         if restrict_input:
             # add sufficient info to compute
             # conformal interval
             main_logger.info("Restricting input...")
-            new_input_size = 3 * self.input_size + max(self.model_horizon, self.h)
+            if self.level is not None:
```
jmoralez, line 567:
```suggestion
            if self.level is None:
```

AzulGarza: thanks. this was intended to be `self.level is not None`. i've tried to explain the reason here: https://github.com/Nixtla/nixtla/blob/603c6669eec58194bbf005fefa0417ec72bfd891/nixtla/nixtla_client.py#L568-L574

basically, when a user passes a large amount of data and only requires prediction intervals (`self.level is not None`), we can set an upper limit on the data sent to the API to reduce the latency by reducing the conformal intervals calibration windows. if a user passes a large amount of data and also an exogenous variable or wants to fine-tune, we send the complete data, regardless of the value of `level`.

AzulGarza: it was an opinionated decision back then. but I've tried to maintain the behavior for the change to be backward compatible. we can, of course, change this and remove it.

jmoralez: I see. So the comment is what was wrong then https://github.com/Nixtla/nixtla/blob/1e0b0a8bf0d86939fb367ee8d88dc49f306f99a3/nixtla/nixtla_client.py#L560

So I should close #265, right?

AzulGarza: oh i see, that was the source of confusion, sorry about that xD fixed in https://github.com/Nixtla/nixtla/pull/314/commits/e1c278903f57494cd5056d2ae41ae591e9afa266

yes, i think we can close it.

jmoralez (approved):

PR #315, opened 2024-04-25
Update openapi json file

AzulGarza:
Created by Github action


AzulGarza:
@jmoralez , it worked!🥳 closing it since only different random values were included (in the future we would probably need to think about what to do when there are changes).

PR #318, opened 2024-04-26
store x_cols as a variable instead of attribute

jmoralez:
Since the x_cols are used only internally in the same method this changes them to be stored in a local variable instead of a class attribute. This helps prevent unwanted interactions when performing multithreading.


AzulGarza (approved):
lgtm :)

PR #321, opened 2024-04-29
Electricity demand use case

marcopeix:
Compared TimeGPT and N-HiTS for a problem of electricity demand forecasting using [this dataset ](https://zenodo.org/records/4659727) from the Monash Repository, based on [this tutorial](https://cienciadedatos.net/documentos/py29-forecasting-electricity-power-demand-python.html) from SkForecast.

TimeGPT is more accurate and faster than N-HiTS.

Include use case for intermittent demand. TimeGPT performs 14% better than the top statistical model for intermittent forecasting.


mergenthaler:
>Not sure how to handle the dataset, since I had to download it locally. Should I add it to the transfer-learning repo and use the GitHub raw link after?
Yes, please.

mergenthaler:
@marcopeix, is this ready to be reviewed by Azul and merged?

marcopeix:
@AzulGarza , ready for review! 😄

nbs/docs/5_use_cases/3_electricity_demand.ipynb
```diff
@@ -0,0 +1,934 @@
+{
```
AzulGarza, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    input_df = df.groupby('unique_id').apply(lambda group: group.iloc[-1104:-96]).reset_index(drop=True)   # 1008 = 21 days (1008 *0.5h * 1 day/24h)
</pre><p>to prevent hardcoding <code>roup.iloc[-1104:-96]</code>,you could use <code>input_df = df.drop(test_df.index).reset_index(drop=True)</code>   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/321/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

AzulGarza, line 1: <pre class="ql-syntax" spellcheck="false">Line #23.    df = read_tsf("data/australian_electricity_demand_dataset.tsf")
</pre><p>i think we need to change this dataframe to <code>df = read_tsf_from_url(url)</code>.also i think we can reuse <code>test_df</code> and <code>input_df</code> for both pipelines to prevent defining the same dataframes twice.     </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/321/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

AzulGarza (changes requested):
thanks @marcopeix! i left a couple of comments :)

nbs/docs/5_use_cases/3_electricity_demand.ipynb
marcopeix: <p>I can't use the entire dataframe, because it is too large and it causes my kernel to crash locally. I need to select part of the data only. I'll reuse the same input and test dfs, but I'll have to slice.</p><p> </p>

marcopeix: <p>Agreed, will fix this!</p>

.github/workflows/ci.yaml
```diff
@@ -65,7 +65,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-13, windows-latest]
```
AzulGarza, line 68: is there something in particular that breaks `macos-latest`? perhaps we can fix it in the notebooks, since the rest of the docu seems to work well with `macos-latest`.

AzulGarza (changes requested):
thanks @marcopeix! 

i left a qq. also, could you please update [mint.json](https://github.com/Nixtla/nixtla/blob/main/nbs/mint.json) including the new notebooks? 
thanks!

.github/workflows/ci.yaml
marcopeix: On local, everything works fine, but in the CI, I encountered this error: 

`RuntimeError: MPS backend out of memory (MPS allocated: 0 bytes, other allocations: 0 bytes, max allowed: 7.93 GB). Tried to allocate 480.00 KB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).`

I reduced the batch size and val_batch_size, but that didn't fix the CI. José suggested I change it to `macos-latest`.

AzulGarza (approved):

PR #323, opened 2024-04-30
[DOCS] New structure

elephaint:
- New structure for TimeGPT documentation


mergenthaler:
@AzulGarza, this is ready for review.

mergenthaler:
Quick question @elephaint (and @AzulGarza ). From what I understand, the order of the documents in the sidebar from readme.com is given by the [slug](https://docs.readme.com/main/reference/updatedoc) flag passed in the [create readme script](https://github.com/Nixtla/nixtla/blob/23861e3a2de19d53d4135e71a7f92d4cee0bb9e5/action_files/readme_com/create_readme_docs.sh#L15).  Right now the order is given by increasing the `counter`. 
My two questions:
Does this mean that we have to make sure the files are ordered correctly in the folders? 
Is this file iteration logic ok or should we explore something different?

mergenthaler:
Makes sense @elephaint. Let's see what @AzulGarza thinks. But if she agrees, I'm ok with merging this PR and creating a new one that deals specifically with ordering nbs inside the subfolders. I think ultimately, being able to define the order makes a lot of sense.

AzulGarza:
thanks @elephaint @mergenthaler! 

> Also, I noticed that in the rendered readme often blank lines in the notebooks are ignored, as well as some bullet lists. So there is still a lot of reformatting to do (but that can be done separately from having the structure in place already).

i agree, i think we should also fix some rendering issues with the images. eg hierarchical:

image

 i'm merging this pr to have the new structure in main and iterating on that.

AzulGarza (approved):

PR #324, opened 2024-04-30
[FIX] Doc fixes

elephaint:
- Render images properly everywhere (checked for readme.com)
- Correct Readme.com order and structure of docs
- Correct line spacing

Please see https://dash.readme.com/project/timegpt/v0.0.2 for examples of how the docs will look like on readme.com.

Open issues / questions @mergenthaler @AzulGarza:
- New image rendering in mintlify not checked yet, as I can't get it to render locally.
- Readme.com still has 'flat' sidebar structure - I can create the nested layout alike Mintlify too, wdyt? (is another bit fiddling with creating section notebooks for all these sections)


mergenthaler:
Thanks @elephaint. I think we should def. create the nested layout. But we can do this in another PR.
If @AzulGarza agrees we should merge this.

AzulGarza (approved):
thanks @elephaint, lgtm :)

PR #330, opened 2024-05-02
Updates to introduction and quickstart

tracykteal:
* Moved content from the Introduction section of Getting Started over to the Welcome/Introduction page
* Created Steps in the QuickStart 
* Added creating the API key and installing nixtla as the first two steps, so people can walk through that if they haven't done that already

See what you think about the Steps idea.


elephaint:
Changes are included in #329

PR #331, opened 2024-05-03
fix: remove prefixes from paths in mint.json

rpmccarter:
Several fielpaths must have changed, and their corresponding nav entries did not. We still need to fix syntax errors on a few pages and add some new files to the nav, but this should fix the majority of pages

### DIFF
--- a/mint.json
+++ b/mint.json
@@ -27,9 +27,10 @@
     {
       "group": "Getting Started",
       "pages": [
-        "docs/1_getting_started/2_quickstart.html",
-        "docs/1_getting_started/3_setting_up_your_api_key.html",
-        "docs/1_getting_started/7_faqs.html"      
+        "docs/1_getting_started/introduction.html",
+        "docs/1_getting_started/quickstart.html",
+        "docs/1_getting_started/setting_up_your_api_key.html",
+        "docs/1_getting_started/faqs.html"      
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


elephaint:
@rpmccarter Thanks - I think this is due to an earlier merge conflict that did not surface. Agree that the fix should work, let me check whether if we merge our latest changes (which include a set of other changes too) this will be fixed automatically too.

jmoralez:
Also this PR should target [this file](https://github.com/Nixtla/nixtla/blob/main/nbs/mint.json) in the main branch.

elephaint:
I think if we merge #329 this will also be fixed

elephaint:
> Several fielpaths must have changed, and their corresponding nav entries did not. We still need to fix syntax errors on a few pages and add some new files to the nav, but this should fix the majority of pages

Out of interest, why are prefixes removed from filenames by Mintlify? Is that defined somewhere? Because it feels quite weird....

elephaint:
@rpmccarter We implemented these changes but it still doesn't render. Can you look at the [logs of build_docs](https://github.com/Nixtla/nixtla/actions/runs/8944367130/job/24571136068) and see what goes wrong? There is no error there, so I'd assume the files are built properly.

mergenthaler:
@elephaint, is this already fixed? If yes, please feel free to close this PR.

elephaint:
Yes, this has been fixed already.

PR #334, opened 2024-05-03
[FIX] Doc fixes

elephaint:
- Fix some image tags, causing many docs not to render on Mintlify
- Fix FAQ for Mintlify
- Fix a few Colab badges
- Remove index from Mintlify for consistency with Readme.com
- Fix some links
- Remove `pip install` as codepart and include it as call-out block in some notebooks
- Auto hide capabilities placeholder notebook


action_files/readme_com/modify_markdown.py
```diff
@@ -65,13 +65,19 @@ def modify_markdown(
     except:
         pass
 
+    # Hide the unnecessary capabilities notebook for readme.com
```
AzulGarza, line 68: nice

AzulGarza (approved):
i had doubts about removing `nbs/index.ipynb`, but we are modifying the readme by hand. so we can remove It.

PR #349, opened 2024-05-08
[FIX] Colab-flag

elephaint:
- Adds Colab flags to all notebooks
- Adds Colab badge to capabilities notebooks


AzulGarza (changes requested):
thank you so much @elephaint, it seems it involved a lot of effort. i'm very sorry about asking a change to his pr, but would you be open to implementing the following logic? i think it can improve readability and might be easier to maintain:

- define `IN_COLAB = 'google.colab' in sys.modules` in `nixtla/utils` so we can import it instead of defining each time.
- import all the required deps for colab in one cell:

```
#| hide
if not IN_COLAB:
    from nixtla.utils import colab_badge
    rom dotenv import load_dotenv

elephaint:
> thank you so much @elephaint, it seems it involved a lot of effort. i'm very sorry about asking a change to his pr, but would you be open to implementing the following logic? i think it can improve readability and might be easier to maintain:
> 
> * define `IN_COLAB = 'google.colab' in sys.modules` in `nixtla/utils` so we can import it instead of defining each time.
> * import all the required deps for colab in one cell:
> 
> ```
> #| hide
> if not IN_COLAB:
>     from nixtla.utils import colab_badge
>     rom dotenv import load_dotenv
> ```

@AzulGarza in Colab nixtla is not installed, so we can't import IN_COLAB when it's part of nixtla. Unless the first line in Colab is `%pip install nixtla', but then we can't make that install conditional on being in Colab.

lightning_logs/version_4/events.out.tfevents.1715239615.ospc
AzulGarza, line 1: @elephaint could you please remove these files? i think they were pushed by mistake. :)

AzulGarza (changes requested):

lightning_logs/version_4/events.out.tfevents.1715239615.ospc
elephaint: Yes, indeed...

AzulGarza (approved):

PR #366, opened 2024-05-16
[Hotfix] Supported DataFrames

MMenchero:
## Description 

This fixes a minor error regarding supported DataFrames in TimeGPT. The original version said that "Currently, TimeGPT only supports pandas DataFrames...". However, when using a distributed framework, TimeGPT also accepts Spark, Dask and Ray DataFrames. 

Changes done in this PR aim to make this clear.


MMenchero:
Made this changes based on your comment @mergenthaler.

PR #386, opened 2024-06-04
hotfix: support azureai model as in the docs

AzulGarza:
this is a hardcoded solution for model=azureai when using an azure endpoint.


nixtla/nixtla_client.py
```diff
@@ -739,6 +740,7 @@ def wrapper(self, *args, **kwargs):
                     f'unsupported model: {kwargs["model"]} '
                     f'supported models: {", ".join(self.supported_models)}'
                 )
+            kwargs["model"] = model
```
jmoralez, line 743: I think this line will make it fail, since the azure deployment expects the model to be azureai and this would send timegpt-1

PR #407, opened 2024-07-04
Update the middleware endpoint to use the new one

loama:


AzulGarza (approved):
lgtm, thank you @loama!

Issue #419, opened 2024-07-10
[Feat] Add Support for UTC  date stamps.


mergenthaler:
Currently, passing dates like `2023-06-02 00:00:00-06:00` results in a `500` error.

mergenthaler:
#429 will fix this :)

PR #133, opened 2023-10-24
[FEAT] Add cross validation method

AzulGarza:
This PR includes the `cross_validation` method and a tutorial on how to use it.

Issue #116, opened 2023-09-29
[FIX] Ensure target variable is numerical.


AzulGarza:
Check that the target variable and exogenous variables are numerical before sending the request.

mergenthaler:
Fixed by #429.

Issue #96, opened 2023-08-29
Token limit (not used, but expired)


wadood-flextock:
Created a basic timeseries to predict. 
Applied the token from the https://dashboard.nixtla.io/
Was out of limit on second run (first was on error)

ApiError: status_code: 429, body: {'data': None, 'message': 'Too many requests', 'details': 'You have reached your request limit, email ops@nixtla.io to continue using the API', 'code': 'A20', 'requestID': 'NGJ6Y5GZ7F', 'support': 'If you have questions or need support, please email ops@nixtla.io'}

BasilaryGroup:
At least you were able to run it once. It now fails before completing the first run. See https://github.com/Nixtla/nixtla/issues/173

TingTingwoaini:
> Created a basic timeseries to predict. Applied the token from the https://dashboard.nixtla.io/ Was out of limit on second run (first was on error)
> 
> ApiError: status_code: 429, body: {'data': None, 'message': 'Too many requests', 'details': 'You have reached your request limit, email [ops@nixtla.io](mailto:ops@nixtla.io) to continue using the API', 'code': 'A20', 'requestID': 'NGJ6Y5GZ7F', 'support': 'If you have questions or need support, please email [ops@nixtla.io](mailto:ops@nixtla.io)'}

i have the same question

bai-lf:
zhe same question

elephaint:
I believe this is likely no longer relevant - [you can set up an API key here](https://nixtlaverse.nixtla.io/nixtla/docs/getting-started/setting_up_your_api_key.html).

PR #420, opened 2024-07-11
[FEAT] add support for UTC timestamps

MMenchero:
## Description 

Addresses issue #419. 

### Status 
- **Single time zone support**: Handles UTC timestamps that include a single, unique time zone, such as `2023-06-02 00:00:00-06:00`.
- **Output consistency**: If the input timestamps contain UTC data, the output will also use the same time zone formatting. 
 
### Limitations
Currently, this does not support inputs that contain multiple time zones.


nbs/nixtla_client.ipynb
```diff
@@ -267,6 +267,8 @@
     "        self.drop_uid: bool = False\n",
```
jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #46.            self.utc_flag: bool = False
</pre><p>We don't need to initialize these here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #119.                    self.timezone = df["ds"].dt.tz
</pre><p>The <code>tz</code> attribute is <code>None</code> if the timestamp doesn't have a timezone, so we don't need two variables, we can just extract tz and check if it's None or not  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #120.                    df["ds"] = df["ds"].astype(str)
</pre><p>We should remove the timezone first to avoid the regex, i.e. <code>df['ds'] = df['ds'].dt.tz_localize(None)</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

MMenchero: removed

MMenchero: I agree, this is more efficient.

MMenchero: ok

nbs/nixtla_client.ipynb
```diff
@@ -267,6 +267,7 @@
...
```
jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #46.            self.timezone: str = None
</pre><p>This can be removed</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #115.            if df.dtypes.ds != "object": # then 'ds' must be datetime 
</pre><p>I think it'd be better to use <code>if pd.is_datetime64_any_dtype(df['ds'].dtype):</code> here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #116.                if df["ds"].dt.tz is not None:
</pre><p>You can unconditionally assign the timezone and later on check if it was None. Same for <code>times.tz_localize(None)</code>, if it doesn't have a time zone it won't do anything, so you can instead do:</p>

<pre class="ql-syntax" spellcheck="false">self.timezone = df["ds"].dt.tz
df["ds"] = df["ds"].dt.tz_localize(None)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 267: <pre class="ql-syntax" spellcheck="false">Line #130.                if X_df.dtypes.ds != "object":
</pre><p>Please change this as well to check if the data type is datetime</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

MMenchero: I was under the impression that this was necessary to prevent the `AttributeError: '_NixtlaClientModel' object has no attribute 'timezone'` from the Models Performance check. 
In any case, should the correct type be `self.timezone: Optional[str] = None`?

jmoralez: You can assign it in the else branch of the if that checks if the ds is datetime

MMenchero: makes sense since it is not always needed.

nbs/nixtla_client.ipynb
```diff
@@ -335,8 +335,13 @@
     "            self.target_col: \"y\",\n",
```
jmoralez, line 335: <pre class="ql-syntax" spellcheck="false">Line #133.                else: 
</pre><p>Please remove this part. It isn't likely but if the training df has a timezone and the future doesn't this would remove it. We should rely on the training one.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jmoralez, line 335: <pre class="ql-syntax" spellcheck="false">Line #7.    fc.head()
</pre><p>It would be better to assert here that the timezone in the output is the same as the one in the input.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='82'/>

jmoralez, line 335: <pre class="ql-syntax" spellcheck="false">Line #7.    df = df[df['ds'].str.endswith("-06:00")]
</pre><p>Why is this being done?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='83'/>

MMenchero: I’m afraid I’m confused about the best approach to do this:

1. I added the
```
else: 
 self.timezone = None 
```
to avoid the `AttributeError` from the Models Performance check in GitHub. However, I can see how it can cause problems if the training df has a timezone but the future df doesn’t. 

2. Earlier on we mentioned that there is no need to initialize `self.timezone` in the constructor. 

3. Hence, I’m not sure where is the best place to define to avoid the error and also keep the code clean since not in all cases the timezone is defined.

MMenchero: makes sense and is a more robust test than just looking at the output.

MMenchero: Because when using `df['ds'] = df['ds'].dt.tz_convert('America/Mexico_City')`, the length of the data makes two timezones appear due to daylight savings time: `-05:00` and `-06:00`. This case is not supported by our current code. 

With this check, I just wanted to show that the current solution works with timezones other than `+00:00`.

jmoralez: We just need the else in the check for the training df, not the future one.

jmoralez: It's the same timezone, just two offsets. The current code handles that correctly when restoring it.

MMenchero: ah! got it.

MMenchero: Indeed, it's the same timezone. However, when running this with the current code: 
```
df = pd.read_csv('https://raw.githubusercontent.com/Nixtla/transfer-learning-time-series/main/datasets/electricity-short.csv')
df['ds'] = pd.to_datetime(df['ds'], utc=True)
df['ds'] = df['ds'].dt.tz_convert('US/Eastern')

fc = nixtla_client.forecast(df, h=8, level=[80,95])
```
I get this error: `Could not infer frequency of ds column. This could be due to inconsistent intervals. Please check your data for missing, duplicated or irregular timestamps`. 

That is why I reduced the dataset to one offset.

jmoralez: Then we should remove the timezone before we infer the frequency

MMenchero: I tried that, but then I noticed the following problem. 

Assume we are converting the dates to US/Eastern. 
```
df = pd.read_csv('https://raw.githubusercontent.com/Nixtla/transfer-learning-time-series/main/datasets/electricity-short.csv')
df["ds"] = pd.to_datetime(df['ds'], utc=True).dt.tz_convert('US/Eastern')
```
When daylight savings occurs, we end up with this data 
```
BE,2016-11-06 00:00:00-04:00,30.1
BE,2016-11-06 01:00:00-04:00,29.82
BE,2016-11-06 01:00:00-05:00,30.89
BE,2016-11-06 02:00:00-05:00,30.89
```
So, when using `df["ds"] = df["ds"].dt.tz_localize(None)`, we end up with duplicates on the day daylight savings starts
```
BE,2016-11-06 00:00:00,30.1
BE,2016-11-06 01:00:00,29.82
BE,2016-11-06 01:00:00,30.89
BE,2016-11-06 02:00:00,30.89
```
And this generates a problem later on, namely `cannot handle a non-unique multi-index!` Still thinking how to fix this...

jmoralez: I see. I think we need to convert to UTC first and then remove the timezone, e.g. `df["ds"].dt.tz_convert('utc').tz_localize(None)`

MMenchero: I did this. I also think it's worth explaining this new line of code: `df = df.drop_duplicates(subset=[self.id_col, self.time_col])`

I noticed that this code: 
```
df = pd.read_csv('https://raw.githubusercontent.com/Nixtla/transfer-learning-time-series/main/datasets/electricity-short.csv')
df["ds"] = pd.to_datetime(df['ds'], utc=True).dt.tz_convert('US/Eastern')
```
generates these values on the day daylight savings starts
```
BE,2016-11-06 00:00:00-04:00,30.1
BE,2016-11-06 01:00:00-04:00,29.82
BE,2016-11-06 01:00:00-05:00,30.89
BE,2016-11-06 02:00:00-05:00,30.89
```
So when we use `df["ds"] = df["ds"].dt.tz_localize(None)`, we end up with duplicates on the day daylight savings starts
```
BE,2016-11-06 00:00:00,30.1
BE,2016-11-06 01:00:00,29.82
BE,2016-11-06 01:00:00,30.89
BE,2016-11-06 02:00:00,30.89
```
This leads to a problem later on, namely `cannot handle a non-unique multi-index! `. The new line of code is meant to prevent this.

jmoralez: That'd loose data. If you convert to utc first you make sure each timestamp is unique and thus removing the timezone just removes the 0 offset.

So the process is TZ aware timestamp (e.g. US/Eastern) -> convert to UTC (this is done without loosing information nor generating duplicates -> turn into naive timestamp.

MMenchero: I missed your earlier comment before I made the commit. I'll try this since it avoids removing duplicated.

nbs/nixtla_client.ipynb
```diff
@@ -317,6 +317,13 @@
     "    def transform_inputs(self, df: pd.DataFrame, X_df: pd.DataFrame):\n",
```
jmoralez, line 317: <pre class="ql-syntax" spellcheck="false">Line #98.                df[self.time_col] = df[self.time_col].dt.tz_convert('utc')
</pre><p>I think this requires an if now, since tz_convert only works for tz aware timestamps, so these two lines should go inside an <code>if self.tz is not None:</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

MMenchero: agree, turns out `tz_convert` results in an error when using the dataset with `df['ds'] = pd.to_datetime(df['ds'])`.

MMenchero: there's now an error with trying to pass a dataframe with index, so let me work on that. This is where the error happens: 
`fcst_inferred_df_index = nixtla_client.forecast(df_ds_index, h=10, time_col='timestamp', target_col='value')`

MMenchero: **Current status**

1. We need to transform the data before inferring the frequency. Otherwise, the UTC part generates an error when attempting to infer the frequency. 

2. We need to consider the case where the time_col is the index. This case is handled after inferring the frequency, but it is too late for the UTC conversion that expects the time_col column.  

3. The current fix might not be the right solution since it could prevent the condition `self.base_freq is None and hasattr(df.index, "freq")` from ever being True, and it doesn't pass the Ubuntu checks.

nbs/nixtla_client.ipynb
```diff
@@ -216,6 +216,13 @@
     "}"
```
AzulGarza, line 216: <pre class="ql-syntax" spellcheck="false">Line #96.            if df.index.name == self.time_col:
</pre><p>what happens if we check the index before updating the timezone?</p><p> </p><p>something like the following. in any case, we reset the index but we are able to pass the frequency in the index.</p>

<pre class="ql-syntax" spellcheck="false">&nbsp; &nbsp; &nbsp;  if self.base_freq is None and hasattr(df.index, "freq"):
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; inferred_freq = df.index.freq
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; if inferred_freq is not None:
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; inferred_freq = inferred_freq.rule_code
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; main_logger.info(f"Inferred freq: {inferred_freq}")
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; self.freq = inferred_freq
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; time_col = df.index.name if df.index.name else "ds"
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; self.time_col = time_col
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; df.index.name = time_col
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; df = df.reset_index()
&nbsp; &nbsp; &nbsp; &nbsp; if df.index.name == self.time_col:
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; df.reset_index(inplace=True)
&nbsp; &nbsp; &nbsp; &nbsp; if pd.api.types.is_datetime64_any_dtype(df[self.time_col].dtype):
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; self.timezone = df[self.time_col].dt.tz
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; if self.timezone is not None:
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; df[self.time_col] = df[self.time_col].dt.tz_convert('utc')
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; df[self.time_col] = df[self.time_col].dt.tz_localize(None)
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; df[self.time_col] = df[self.time_col].astype(str)
&nbsp; &nbsp; &nbsp; &nbsp; else:
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; self.timezone = None
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/420/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

MMenchero: <p>Thanks for the suggestion @AzulGarza. This passes our checks for timezone-aware timestamps. Unfortunately, it then generates an error in the <code>distributed.nixtla_client</code> notebook. Namely, when running the following tests for Dask and Ray</p><p> </p><blockquote>test_forecast_x_dataframe(dask_df_x, dask_future_ex_vars_df)</blockquote><blockquote>test_forecast_x_dataframe(ray_df_x, ray_future_ex_vars_df)</blockquote><p> </p><p>it raises the following error: <code>cannot infer freq from a non-convertible dtype on a Series of string</code></p><p>This error comes from when we try to infer the frequency, namely this line </p><p> </p><blockquote>Line 174. 		inferred_freq = pd.infer_freq(df_id['ds'].sort_values())</blockquote><p> </p><p>Still trying to come up with a solution for this... </p>

AzulGarza:
@jmoralez @MMenchero should we close this since #429 was merged?

PR #436, opened 2024-07-26
[DOC] - Rephrase the sentence for specific loss function

marcopeix:
Right now, it says that we can use a custom loss function. That may lead our users to think that they can define their own loss function, when in reality, they can choose from a defined list.


AzulGarza (approved):
thank you, @marcopeix!💙

PR #443, opened 2024-08-07
feat: use TypeVar for dataframes

jmoralez:
Changes the type hints for the input and output dataframes from using a type union to a typevar. The main benefit of this is that this tells the type checkers that whatever the input of a function was (restricted to some dataframe types) the output will be of the same type, which helps a lot with IDES and type checkers, for example:

Pandas:
Screenshot from 2024-08-16 13-41-54

Polars:
Screenshot from 2024-08-16 13-42-25

Also packages the `py.typed` file, which was already there but wasn't being included in the distribution. This tells type checkers that the package has type annotations.


setup.py
```diff
@@ -34,6 +34,8 @@
     long_description_content_type="text/markdown",
     url="https://github.com/Nixtla/nixtla",
     packages=setuptools.find_packages(exclude=["action_files"]),
+    include_package_data=True,
+    package_data={'nixtla': ['py.typed']},
```
AzulGarza, line 38: cool

AzulGarza (approved):

setup.py
jmoralez: Seems that I lied haha. It's already being packaged ([0.5.2 logs](https://github.com/Nixtla/nixtla/actions/runs/9807622855/job/27081741074#step:6:226)), I'll remove this.

PR #449, opened 2024-08-19
[DOCS] Add API call count FAQ

marcopeix:
Add a section if FAQ to describe what counts as an API call. 

Covers situations with cross-validation, `num_partitions`, and forecast.


nbs/docs/getting-started/5_faq.ipynb
```diff
@@ -470,6 +470,24 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<details>\n",
+    "    <summary>What counts as an API call?</summary> \n",
+    "\n",
+    "An API call is a request made to TimeGPT to perform some an action like forecasting or detecting anomalies.\n",
```
elephaint, line 480: Typo in this sentence, `to perform an action`

nbs/docs/getting-started/5_faq.ipynb
```diff
@@ -470,6 +470,24 @@
...
+    "\n",
+    "For example, if we pass an entire dataset and call the forecast method once, then it counts as 1 API call.\n",
+    "\n",
+    "In some cases, the dataset can be too large and the API will return an error. Then, we must partition the data, using the `num_partitions` parameter (see our tutorial on [computing at scale](https://docs.nixtla.io/docs/tutorials-dask)). For example, we set `num_partitions=2`, then there will be 2 API calls.\n",
+    "\n",
+    "If we perform [cross-validation](https://docs.nixtla.io/docs/tutorials-cross_validation) using five windows, then it counts as 5 API calls, since the model must perform the forecast action twice (once per window).\n",
```
elephaint, line 486: This confuses me - first we state five windows == five calls, then we talk about the forecasting action being performed twice (once per window)? 

I don't really get it?

elephaint (changes requested):
Nice work! See comments

nbs/docs/getting-started/5_faq.ipynb
MMenchero: I would also remove the link to the computing at scale tutorials or put it at the end because those tutorials are about how to use `TimeGPT` with very specific frameworks and not really what `num_partitions` is about.

marcopeix: Looks like on that day, 2 and 5 were the same for me! I fixed the typos and the confusion. I agree with removing the link to distirbuted computing.

nbs/docs/getting-started/5_faq.ipynb
```diff
@@ -470,6 +470,24 @@
...
+   "metadata": {},
+   "source": [
+    "<details>\n",
+    "    <summary>What counts as an API call?</summary> \n",
+    "\n",
+    "An API call is a request made to TimeGPT to perform an action like forecasting or detecting anomalies.\n",
+    "\n",
+    "For example, if we pass an entire dataset and call the forecast method once, then it counts as 1 API call.\n",
+    "\n",
+    "In some cases, the dataset can be too large and the API will return an error. Then, we must partition the data, using the `num_partitions` parameter. For example, if we set `num_partitions=2`, then there will be 2 API calls.\n",
```
elephaint, line 484: I think we should mention the dataset size limit here (200MB right?), so users know how big their data can be (roughly) before it needs to be partitioned. E.g.

```
If the API call requires to send more than 200MB of data, the API will return an error and will require you to use the `num_partitions` parameter in order to partition your request. Every partition will count as an API call, thus for 2 partitions there will be 2 API calls.
```

elephaint (changes requested):
I think we should mention the dataset size limit before we partition, so that users have a rough idea on how large they can go before they need to partition.

nbs/docs/getting-started/5_faq.ipynb
marcopeix: Good point!

tracykteal:
Is it just 1 API call per forecast? Currently we need to make two requests to the API to generate forecasts. The first one gets the model's parameters to perform some preprocessing on the data, and the second one calls the actual path. 

So, when people make one forecast, they see 2 API requests in their dashboard. This could be a place where we explain that too.

elephaint (approved):
It needs a few reads to understand, but that's it's because complex how we made it (ie. the API call structure)

LGTM!

PR #452, opened 2024-08-21
feat: orjson serialization

jmoralez:
Uses [orjson ](https://github.com/ijl/orjson) to serialize the payloads and to deserialize the responses. This is considerably faster because we don't need to convert the arrays to lists and then use `json.dumps` (which is slower than `orjson.dumps` by itself), we just provide the arrays directly to orjson. Parsing the responses is also faster.


nixtla/nixtla_client.py
```diff
@@ -1298,7 +1307,14 @@ def cross_validation(
             time_col=time_col,
             target_col=target_col,
         )
-        targets = df[target_col].to_numpy()
+        if isinstance(df, pd.DataFrame):
+            # in pandas<2.2 to_numpy can lead to an object array if
+            # the type is a pandas nullable type, e.g. pd.Float64Dtype
+            # we thus use the dtype's type as the target dtype
+            target_dtype = df.dtypes[target_col].type
```
AzulGarza, line 1314: omg, love the deep understanding of this case

AzulGarza (approved):
awesome, @jmoralez! lgtm

PR #453, opened 2024-08-22
[FEAT] Add historical exogenous

elephaint:
- Adds possibility to use historical exogenous variables. Basically, historical exogenous variables are those variables in `df` that are not in `X_df`
- Updates docs to show how historical exogenous variable can be used


jmoralez (approved):
LGTM on the implementation. Maybe @mergenthaler has some SEO tricks up his sleeve.

PR #454, opened 2024-08-26
[DOC] - Add instructions on saving figures when not working in notebooks

marcopeix:
Add a callout to avoid [this issue](https://docs.nixtla.io/discuss/669d8327ca6cb300121fdad7). Shows how to save figures when working outside of notebooks.


elephaint (approved):
LGTM!

PR #456, opened 2024-08-28
[FEAT] - Raise error when payload is too large and suggest number of partitions

marcopeix:
When the payload exceeds 200MB, an error is raised and we suggest the minimum number of partitions to be used.

For example, a payload of 275MB raises: `ValueError: The payload is too large. Set num_partitions=2`.


nixtla/nixtla_client.py
```diff
@@ -620,6 +621,11 @@ def ensure_contiguous_arrays(d: Dict[str, Any]) -> None:
 
         ensure_contiguous_arrays(payload)
         content = orjson.dumps(payload, option=orjson.OPT_SERIALIZE_NUMPY)
+        content_size_mb = sys.getsizeof(content) / (1024 * 1024)
```
jmoralez, line 624: `content` is already a bytes object, we just want to count the number of bytes, not get the size of the python object.
```suggestion
        content_size_mb = len(content) / (1024 * 1024)
```

jmoralez (changes requested):
Please also clean the notebook with `nbdev_clean --fname nbs/nixtla_client.ipynb --clear_all`

jmoralez (approved):

PR #457, opened 2024-08-28
feat: suggest earlier version for azure endpoints

jmoralez:


nixtla/nixtla_client.py
```diff
@@ -601,6 +601,15 @@ def __init__(
...
+            from packaging.version import Version
+
+            import nixtla
+
+            if Version(nixtla.__version__) > Version("0.5.2"):
+                raise NotImplementedError(
+                    "This version doesn't support Azure endpoints, please install "
+                    "an earlier version with: `pip install 'nixtla<=0.5.2'`"
+                )
             self.supported_models = ["azureai", "timegpt-1-long-horizon"]
```
jmoralez, line 613: I'd originally done this to support older azure deployments, but I think this should now just be `["azureai"]`, right @AzulGarza?

AzulGarza: yes, exactly, thank you @jmoralez

AzulGarza (approved):
awesome, @jmoralez!🎉

PR #471, opened 2024-09-10
FEAT: Add finetune_depth parameter

marcopeix:
Add the `finetune_depth` parameter to control how many layers are finetuned.
Adjust tutorials and capabilities with new parameter.


nixtla/nixtla_client.py
```diff
@@ -898,6 +900,9 @@ def forecast(
         finetune_steps : int (default=0)
             Number of steps used to finetune learning TimeGPT in the
             new data.
+        finetune_depth: int (default=1)
+            Depth of finetuning, on a scale from 1 to 5. 1 means only the last layer is finetuned
```
elephaint, line 904: We shouldn't talk about layers in the finetuning.

See the finetune_depth defintion in api-ltm

nixtla/nixtla_client.py
```diff
@@ -1335,6 +1343,9 @@ def cross_validation(
         finetune_steps : int (default=0)
             Number of steps used to finetune TimeGPT in the
             new data.
+        finetune_depth: int (default=1)
+            Depth of finetuning, on a scale from 1 to 5. 1 means only the last layer is finetuned
```
elephaint, line 1347: See other comment, we shouldn't mention layers.

nbs/docs/capabilities/forecast/07_finetuning.ipynb
```diff
@@ -166,6 +142,35 @@
     "> By default, `timegpt-1` is used. Please see [this tutorial](https://docs.nixtla.io/docs/tutorials-long_horizon_forecasting) on how and when to use `timegpt-1-long-horizon`."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "By default, only the last layer of the model is finetuned. We can finetune more layers using `finetune_depth` parameter. "
```
elephaint, line 149: Not mention layers or any reference to architecture

elephaint (changes requested):
Nice! Few comments, and I think we should include a test in `nixtla_client.ipynb`.

nbs/docs/tutorials/06_finetuning.ipynb
```diff
@@ -326,7 +326,7 @@
     "### 3.1 Fine-tuning more layers with `finetune_depth`\n",
     "It is also possible to control the depth of fine-tuning with the `finetune_depth` parameter.\n",
     "\n",
-    "`finetune_depth` takes values among `[1, 2, 3, 4, 5]`. By default, it is set to 1, which means that only the last layer of the model is fine-tuned. When set to 5, the entire model is fine-tuned. Therefore, `finetune_depth` does not equate to the number of layers being fine-tuned, but it represents a scale of how many layers are fine-tuned, from less to more."
+    "`finetune_depth` takes values among `[1, 2, 3, 4, 5]`. By default, it is set to 1, which means that fine-tuning is shallow. When set to 5, the entire model is fine-tuned. Therefore, `finetune_depth` does not equate to the number of layers being fine-tuned, but it represents a scale of how many layers are fine-tuned, from less to more."
```
elephaint, line 329: We should not mention layers at all. Suggestion:

> `finetune_depth` takes values among `[1, 2, 3, 4, 5]`. By default, it is set to 1, which means mild finetuning, whereas a value of 5 means the maximum amount of finetuning. Increasing `finetune_depth` increases the time to generate predictions.`

nbs/src/nixtla_client.ipynb
```diff
@@ -967,8 +967,8 @@
     "            Number of steps used to finetune learning TimeGPT in the\n",
     "            new data.\n",
     "        finetune_depth: int (default=1)\n",
-    "            Depth of finetuning, on a scale from 1 to 5. 1 means only the last layer is finetuned\n",
-    "            and 5 means the whole model if finetuned.\n",
+    "            The depth of the finetuning. Uses a scale from 1 to 5, where 1 means little finetuning,\n",
```
elephaint, line 970: Nice!

elephaint (changes requested):

nbs/docs/tutorials/06_finetuning.ipynb
marcopeix: It's harder than I though to not mention layers for this parameter! I kept "shallow" instead of "mild", I feel that "mild" confused me more, but I took the rest of the suggestion. Thanks for the help!

elephaint (approved):
LGTM, I removed last couple of mentions of layers.

nixtla/nixtla_client.py
```diff
@@ -1335,6 +1343,9 @@ def cross_validation(
         finetune_steps : int (default=0)
             Number of steps used to finetune TimeGPT in the
             new data.
+        finetune_depth: int (default=1)
+            The depth of the finetuning. Uses a scale from 1 to 5, where 1 means little finetuning,
+            and 5 means that the entire model is finetuned. By default, the value is set to 1.
```
jmoralez, line 1348:
```suggestion
            and 5 means that the entire model is finetuned.
```

nixtla/nixtla_client.py
```diff
@@ -898,6 +900,9 @@ def forecast(
         finetune_steps : int (default=0)
             Number of steps used to finetune learning TimeGPT in the
             new data.
+        finetune_depth: int (default=1)
+            The depth of the finetuning. Uses a scale from 1 to 5, where 1 means little finetuning,
+            and 5 means that the entire model is finetuned. By default, the value is set to 1.
```
jmoralez, line 905:
```suggestion
            and 5 means that the entire model is finetuned.
```

jmoralez (changes requested):
Please also add a test verifying that as the finetune depth increases the loss is lower in the `nbs/docs/tutorials/06_finetuning.ipynb` notebook.

elephaint:
> Please also add a test verifying that as the finetune depth increases the loss is lower in the `nbs/docs/tutorials/06_finetuning.ipynb` notebook.

Such a test doesn't always work; i.e. it's not always the case that finetuning improves the model. So I'm removing it again....

jmoralez:
What changed between now and 7bc1b5d1e554b1ad8fc7c420bcad1b87151b9048? That one has very different results ([nb link](https://github.com/Nixtla/nixtla/blob/7bc1b5d1e554b1ad8fc7c420bcad1b87151b9048/nbs/docs/tutorials/06_finetuning.ipynb)). This is supposed to be deterministic, isn't it? I'd expect to be able to reproduce the metrics from that commit every time, especially the monotonic part, right now 2 and 3 yield the same result which is highly suspicious.
image
image

elephaint:
> What changed between now and [7bc1b5d](https://github.com/Nixtla/nixtla/commit/7bc1b5d1e554b1ad8fc7c420bcad1b87151b9048)? That one has very different results ([nb link](https://github.com/Nixtla/nixtla/blob/7bc1b5d1e554b1ad8fc7c420bcad1b87151b9048/nbs/docs/tutorials/06_finetuning.ipynb)). This is supposed to be deterministic, isn't it? I'd expect to be able to reproduce the metrics from that commit every time, especially the monotonic part, right now 2 and 3 yield the same result which is highly suspicious. image image

Nothing really changed, the issue is that it doesn't hold in general that:
- Finetune depth higher -> better performence
- Finetune steps more -> better performance

I tweaked the parameters so that the results are not good but monotonic (however as said before, finetuning isn't guaranteed to provide results that are strictly better when increasing the parameters)

jmoralez:
Thanks! So the test would pass now? It'd be great having `loss_depth1 > loss_depth2 > loss_depth3` to detect possible regressions or the parameter not being passed through correctly.

elephaint:
> Thanks! So the test would pass now? It'd be great having `loss_depth1 > loss_depth2 > loss_depth3` to detect possible regressions or the parameter not being passed through correctly.

No, because:
- Finetune depth higher does not lead to better performance in general
- Finetune steps higher does not lead to better performance in general
so we shouldn't market that view, either. And a test on that is useless too; if it fails, the results might still be better than before. 

I've updated the example to explain that also (so users see that increasing depth can also worsen performance, and it's a bit of trial and error).

The tutorial fails if the parameter isn't passed through correctly, so we're covered there anyways.

nixtla/nixtla_client.py
jmoralez: This is still here

jmoralez (approved):

nbs/docs/tutorials/06_finetuning.ipynb
```diff
@@ -107,7 +107,9 @@
    "outputs": [],
```
cchallu, line 107: <p>We need to rephrase this so that it is more clear for the user.</p><p> </p><p>Instead of "small amount of fine-tuning is applied", we should phrase it in terms of the number of parameters&nbsp;being fine-tuned. Something like: "<code>finetune_depth</code> ranges from 1 to 5 and controls the number of parameters of the model being fine-tuned. By default, it is set to 1, adjusting a small set of parameters, while a value of 5 fine-tunes the maximum number of parameters. Increasing <code>finetune_depth</code> will also extend the time needed to generate predictions."</p><p> </p><p>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/471/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

cchallu, line 107: <p>I like the conclusion! Some suggestions:</p><p> </p><ul><li>" but it can make it worse too" -&gt; " but it can make it worse too due to overfitting".</li><li>as a tip, mention that higher depth should work better for larger datasets. Relate to previous point saying that this is a 1 series dataset.</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/471/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

cchallu (dismissed):
see notebook

elephaint (approved):
LGTM, I only fixed some typos in the latest revision

PR #473, opened 2024-09-19
adding Azure getting started docs

tracykteal:
Adding a Getting Started for using TimeGEN-1 on Azure


jmoralez:
Hey @tracykteal. Can you please open the PR from a branch within the repo instead of a fork? We set some secrets in there that aren't available in forks.

PR #474, opened 2024-09-19
Adding Azure getting started doc

tracykteal:
Adding a Getting Started guide for working starting with TimeGEN-1 on Azure


nbs/docs/getting-started/22_azure_quickstart.ipynb
```diff
@@ -0,0 +1,660 @@
+{
```
AzulGarza, line 1: <p>we can remove fixing the version. msft deployed v2, and @jmoralez is working on adapting the sdk. 🎉</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/474/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

tracykteal: Thanks! I updated the install instructions to remove the version information. 

I moved the images to a more relevant location in the instructions, but still figuring out how they can appear in the preview correctly.

jmoralez: Why are we not using markdown everywhere? That works in PRs, the main branch and the mintlify docs, does it not work with readthedocs? html tags with absolute links seems to be just asking for trouble

tracykteal:
The images do work for me locally. Since we need to put in the full path for the images, but the images aren't in that path yet, because they're a part of the PR, is there a way to get them to render correctly for the preview? Or do I not have the paths correct for when they're on the site?

Thanks so much!

nbs/docs/getting-started/22_azure_quickstart.ipynb
tracykteal: Agree on markdown! I thought it was an issue with readme.com though?

PR #478, opened 2024-09-25
feat: support newer azure deployments

jmoralez:
The current Azure deployment doesn't support `feature_contributions`. This adds a check to the response to see if the user requested them but they weren't returned and issues a warning if the base url is an Azure one.

Also:
* Makes the methods override the `model` argument when using an Azure deployment, since the only supported model there is `azureai`.
* Adds a missing log message at the start of some methods (Validating inputs).


nixtla/nixtla_client.py
```diff
@@ -1201,6 +1206,8 @@ def detect_anomalies(
                 num_partitions=num_partitions,
             )
         self.__dict__.pop("weights_x", None)
+        model = self._maybe_override_model(model)
+        logger.info("Validating inputs...")
```
AzulGarza, line 1210: cool

nixtla/nixtla_client.py
```diff
@@ -737,6 +728,11 @@ def _make_partitioned_requests(
             ).T
         return resp
 
+    def _maybe_override_model(self, model: str) -> str:
+        if self._is_azure:
+            model = "azureai"
+        return model
+
```
AzulGarza, line 735: nice

AzulGarza (approved):

PR #487, opened 2024-10-03
[FIX] Reduce test time

elephaint:
Reduces some of the test times by:
- Directly loading the filtered data of M5 & LongHorizon for the notebooks that use those datasets.
- Not evaluation code in the test, but evaluating the same piece of code (hidden) for fewer steps at the bottom.


.github/workflows/ci.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
         run: pip install uv && uv pip install --system ".[dev,distributed]"
 
       - name: Run tests
-        run: nbdev_test --timing --do_print --n_workers 0 --flags 'distributed'
```
jmoralez, line 62: This is here to avoid overloading our infra with the CI, please restore it

nixtla/nixtla_client.py
```diff
@@ -664,12 +665,21 @@ def _make_request_with_retries(
...
         endpoint: str,
         payload: Dict[str, Any],
+        lock: Union[None, Lock] = None,
     ) -> Dict[str, Any]:
-        return self._retry_strategy(self._make_request)(
-            client=client,
-            endpoint=endpoint,
-            payload=payload,
-        )
+        if lock is not None:
```
jmoralez, line 670: What's the purpose of this? You do realize this kills the parallelism and would make the one billion experiment run in over 60 hours instead of 6, right?

jmoralez (changes requested):

nixtla/nixtla_client.py
elephaint: Haha, yes I had already removed it

.github/workflows/ci.yaml
elephaint: Easiest way to speed up the tests, though....

jmoralez (approved):

PR #497, opened 2024-10-16
DOCS: Add finetune_depth to tutorial on improving accuracy

marcopeix:
Showcase the `finetune_depth` parameter which controls how many parameters in the model get fine-tuned in the tutorial on improving forecast accuracy with TimeGPT.


AzulGarza (approved):
thank you, @marcopeix! lgtm:)

it's interesting that in this case fine-tuning more parameters leads to similar performance than adding exogenous variables.

PR #500, opened 2024-10-17
updating Azure and trial information

tracykteal:
Updating free trial links to dashboard.nixtla.io, rather than dashboard.nixtla.io/freetrial and related FAQs and some Azure getting-started updates.


carla-cr (approved):
Looks good to me!

tracykteal:
Fixed some links and reviewed live. Set to merge.

jmoralez:
Who reviewed the azure quickstart? The only supported model should be azureai, which is automatically set if an azure endpoint is detected, so there's no need to set it. Also azureai is the same as timegpt-1-long-horizon, so the sections about shorter and longer forecasts should be removed.

tracykteal:
Thanks so much for flagging this @jmoralez. Sorry, things were working for me when I set the model as 'azureai' and wasn't working when I wasn't. I'll re-check my setup to make sure it's working without it, and take out the short and longer forecasts section.

jmoralez:
Were you using a real azure endpoint? I think the override doesn't warn, so behind the scenes we change anything you provided as `model` and set it to `azureai`, so you wouldn't see anything, we probably need to add a warning there.

Can you also add `#| eval: false` to the cells in that tutorial? We run the tests against our own deployment so the CI is failing with errors like this ([sample logs](https://github.com/Nixtla/nixtla/actions/runs/11506997194/job/32032196496#step:5:52)):

> ValueError: unsupported model: azureai. supported models: ['timegpt-1', 'timegpt-1-long-horizon']

tracykteal:
I thought I was using a real azure endpoint, but maybe that's the issue. I'll check with you on what I'm using, and see if I should be using something else. 

Thanks, yes, I'll add in that `#| eval: false`. I'll put that in as its own PR, so the tests will start passing.

PR #532, opened 2024-11-05
feat: support custom and integer frequencies

jmoralez:
Adds support for providing custom pandas offsets as `freq` as well as integer timestamps and `freq=some_int`.


nixtla/nixtla_client.py
```diff
@@ -208,21 +210,29 @@ def _maybe_infer_freq(
...
+        season_length = season_lengths[np.argmax(counts)].item()
+        if season_length <= 7:
+            freq = "D"
+        elif season_length <= 12:
+            freq = "M"
+        elif season_length <= 24:
+            freq = "H"
+        else:
+            freq = "W"
+    return freq
```
AzulGarza, line 235: nice! should we add a warning telling users bout this behavior? wdyt?

nixtla/nixtla_client.py
```diff
@@ -208,21 +210,29 @@ def _maybe_infer_freq(
...
-def _standardize_freq(freq: str) -> str:
-    return freq.replace("mo", "MS")
+def _standardize_freq(freq: _Freq, processed: ufp.ProcessedDF) -> str:
+    if isinstance(freq, str):
+        freq = freq.replace("mo", "MS")
+    else:
+        from coreforecast.grouped_array import GroupedArray
+
+        ga = GroupedArray(processed.data[:, 0], processed.indptr)
+        series_season_lengths = ga._greatest_autocovariance(54).astype("uint8")
```
elephaint, line 224: Isn't this (potentially) slow to do this across all series? Or is `processed.data[:, 0]` only the first?

nixtla/nixtla_client.py
```diff
@@ -208,21 +210,29 @@ def _maybe_infer_freq(
...
+        season_lengths, counts = np.unique(series_season_lengths, return_counts=True)
+        season_length = season_lengths[np.argmax(counts)].item()
+        if season_length <= 7:
```
elephaint, line 227: All in favour of simple business logic here, but why not `<=8`, `<=13`, `<=25`? (To account for situations where autocovariance is slightly higher on the freq +- 1)? (I don't have an experiment to back this up tbh)

nixtla/nixtla_client.py
```diff
@@ -208,21 +210,29 @@ def _maybe_infer_freq(
...
     logger.info(f"Inferred freq: {inferred_freq}")
     return inferred_freq
 
 
-def _standardize_freq(freq: str) -> str:
-    return freq.replace("mo", "MS")
+def _standardize_freq(freq: _Freq, processed: ufp.ProcessedDF) -> str:
+    if isinstance(freq, str):
+        freq = freq.replace("mo", "MS")
+    else:
```
elephaint, line 220: So we run this in case of both integer-freq and pandas-offset-freq? 

In case of integer freq, we're basically saying that the user isn't capable of providing the correct number, right? Because we could just use their integer number instead of the calculated max autocovariance in your business logic below, no?

For the Pandas Offset alias, why not use:
- Match `freq.freq._prefix` to a model we have.
- If no match, do the autocovariance logic.

elephaint (commented):
Unsollicited comments 🙈 

Otherwise looks fine to me

nixtla/nixtla_client.py
jmoralez: This takes 300ms on 10,000 series of lengths between 100 and 1000, so I'd say it's ok. We can improve it further by using multithreading, I'll make that change.

nixtla/nixtla_client.py
jmoralez: This is more related to the input size and the horizon, I don't think it'll make that big of a difference and it's easier to reason about these limits.

nixtla/nixtla_client.py
jmoralez: Yeah this is because we're currently coupling the timestamp frequency with the seasonality. We could also introduce a `season_length` argument and only try to infer it if it's not provided. That'll also help ensuring reproducibility if the user provides less series and that somehow changes the chosen frequency.

I don't follow the `freq.freq._prefix` thing, can you provide an example? `pd.tseries.offsets.CustomBusinessHour` and `pd.tseries.offsets.CustomBusinessDay` don't seem to have that.

elephaint: Maybe to fully understand: this logic is only used to select which of our models (seasonalities) the user ends up using, right?

Then it seems:
- Freq as str -> we match the string freq based on a lookup table to a particular model;
- Freq as int -> we match their given integer number using the bucket logic to a particular model;
- Freq as Pandas offset -> `pd.tseries.offsets.CustomBusinessDay._prefix` == `C` -> we match the prefixes to a particular model using a lookup table.

I don't really see why we need the autocovar logic at all (?). Why not just use a single lookup table that matches strings to the 5ish models we have? That also gives reproducable results, as long as the user provides the same frequency integer / int / pd.offset.

nixtla/nixtla_client.py
elephaint: Now do the 6bln experiment 😛

AzulGarza (approved):

PR #534, opened 2024-11-07
feat: add hist_exog_list argument to cross_validation

### LINKED ISSUE #522: feat: distinguish between historical and future exogenous variables in cross validation
### Description

currently, cross-validation is designed to treat all exogenous variables as future variables by default. moving forward, we aim to align cross-validation settings more closely with forecast settings, enabling differentiation between historical and future exogenous variables.

### Use case

_No response_

jmoralez:
Fixes #522


nbs/src/nixtla_client.ipynb
```diff
@@ -2205,6 +2250,48 @@
...
+    "        train,\n",
+    "        h=h,\n",
+    "        freq=freq,\n",
+    "        hist_exog_list=hist_exog_list,\n",
+    "        X_df=valid,\n",
+    "    )\n",
+    "    np.testing.assert_allclose(\n",
+    "        cv_res['TimeGPT'], fcst_res['TimeGPT'], atol=1e-4, rtol=1e-3\n",
+    "    )"
+   ]
```
AzulGarza, line 2293: cool

AzulGarza: should we add a test to verify that different historical exogenous variables result in different forecasts? wdyt?

jmoralez: added in 91b27ec

AzulGarza (approved):

PR #544, opened 2024-11-22
[FIX] Pricing links incorrect

elephaint:
- Fixes wrong pricing links in FAQ
- Updates contributing with how to handle linking to files.


ngupta23 (approved):

tracykteal (approved):
Thank you for the fix! And updating the CONTRIBUTING documentation, so we get this correct in the future.

PR #555, opened 2024-12-05
[FIX] Electricity short

elephaint:
- Rerun notebooks with electricity _short_ (the previous was ran with the long version, a faux pas)


tracykteal (approved):
Thanks, good idea. This will be easier for others to work with.

PR #556, opened 2024-12-05
feat: add refit argument to cross_validation

jmoralez:
Adds the `refit` argument to `NixtlaClient.cross_validation` to control whether to finetune on all windows or just the first one. This will help speeding up the cross validation process.


AzulGarza (approved):
cool! lgtm

PR #557, opened 2024-12-05
[FIX] Img link

elephaint:
Fixes img link in tutorial why TimeGPT


tracykteal (approved):
Thanks so much @elephaint

PR #558, opened 2024-12-05
[FIX] Img links

elephaint:
Image links to absolute address otherwise they don't work in Mintlify.


jmoralez:
Mintlify works if you use markdown. Why do these have to be HTML tags? Is it because of readme.com?

elephaint:
> Mintlify works if you use markdown. Why do these have to be HTML tags? Is it because of readme.com?

I think this was the only solution that worked without issues on both platforms.

jmoralez:
So yeah, because of readme

jmoralez (approved):

PR #547, opened 2024-11-27
DOC: finetune_depth has its won separate tutorial

marcopeix:
Make a standalone tutorial for `finetune_depth`.
Remove the `finetune_depth` portion of the fine-tuning tutorial.

PR #559, opened 2024-12-05
feat: zstd-compress requests over 1MB

jmoralez:
Automatically compresses payloads over 1MB using [zstandard](https://github.com/facebook/zstd). This should help reduce the time when sending large payloads over slow internet connections.


setup.py
```diff
@@ -48,6 +48,7 @@
         "tenacity",
         "tqdm",
         "utilsforecast>=0.2.8",
+        "zstandard",
```
elephaint, line 51: Should we also set `httpx[zstd]` to ensure correct version is installed?

jmoralez: Oh it seems httpx automatically adds `accept-encoding: 'zstd'` to the headers if zstandard is installed and modal automatically compresses the responses when it sees this, so it's indeed worth adding. Thanks!

setup.py
```diff
@@ -42,8 +42,7 @@
     python_requires=">=3.9",
     install_requires=[
         "annotated-types",
-        "coreforecast>=0.0.14",
```
jmoralez, line 45: This was added in a bad merge in #547

nbs/src/nixtla_client.ipynb
```diff
@@ -2249,6 +2267,48 @@
...
+    "def capture_request():\n",
+    "    original_client = httpx.Client\n",
+    "    httpx.Client = CapturingClient\n",
+    "    try:\n",
+    "        yield\n",
+    "    finally:\n",
+    "        httpx.Client = original_client\n",
+    "\n",
+    "# this produces a 1MB payload\n",
+    "series = generate_series(250, n_static_features=2)\n",
```
elephaint, line 2302: I thought it only compresses above 1 mb? So a 1 mb payload shouldn't compress?

elephaint (approved):
LGTM, two insignificant comments:

- Test is with 1 MB but I thought the idea that it doesn't compress up to (and including) 1 MB?
- Out of curiosity why 1MB as threshold?

jmoralez:
I meant over 1MB in the comment, it's just slightly above it.

The threshold is because at 1MB it takes around 1ms to compress and it's most likely worth it due to the savings in bandwidth, since it always compresses to less than half the original size.

PR #562, opened 2024-12-09
feat: saving and re-using fine-tuned models

jmoralez:
Adds the following:
* `NixtlaClient.finetune` to fine tune and save a model.
* `NixtlaClient.finetuned_models` to list the models a user has fine tuned.
* `NixtlaClient.delete_finetuned_model` to delete an existing model.
* The `finetuned_model_id` argument to all methods to use a saved fine tuned model to compute forecasts.


AzulGarza (commented):
very cool pr! 

when using distributed frameworks, we would expect an error. should we raise a particular error in that case?

jmoralez:
Trying to fine tune on a distributed df?

marcopeix:
Unsolicited comments/questions:

- Why is `num_partitions` not supported for the `finetune` method? Would this limit the finetuning process on small-ish datasets?
- Is it relevant to add a test that compares finetuning in the `forecast` method and finetuning separately and using the finetuned model to ensure the predictions are the same?

jmoralez:
* Supporting partitioning would fine tune `num_partitions` models on unknown portions of the original data. I don't think that's a good use case.
* We already have that test in the API.

AzulGarza (approved):
lgtm:)

elephaint (approved):
Nice, pretty awesome! `output_model_id` (and thus later on `finetuned_model_id`) can be any user chosen string, right? I suspect users will typically want to name their models with a custom string. I think we'll have to emphasize that in the docs following this PR

jmoralez:
> output_model_id (and thus later on finetuned_model_id) can be any user chosen string, right?

Yes, if provided that one is used, otherwise we generate a random UUID.

farukalamai:
I am finetuning the model using my custom dataset. Now how can I save the model like in pkl file or joblib? Here is my code

`
forecast_df = nixtla_client.forecast(
    df=train_df,
    X_df=future_ex_vars_df,
    h=len(future_ex_vars_df),
    id_col='unique_id',
    target_col='y',
    time_col='ds'
)
`

jmoralez:
@farukalamai https://nixtlaverse.nixtla.io/nixtla/docs/tutorials/reusing_finetuned_models.html

PR #563, opened 2024-12-11
feat: add vn1 competition experiment

AzulGarza:
this pr adds an experiment showing that TimeGPT achieves 2nd place in the vn1 forecasting accuracy challenge


MMenchero (commented):
LGTM. Will include an R version.

jmoralez (approved):
I think the R code should also be in nixtlar, or maybe just there.

ngupta23 (approved):

PR #570, opened 2024-12-16
[DOCS] Fixes two docs issues

elephaint:
- Fixes hierarchical forecasting example to reflect changes made to the API of HierarchicalForecast.
- Fixes pricing not appearing on Mintlify by correcting the link to support.


jmoralez (approved):
There can only be one label per PR, so I removed the fix one.

PR #573, opened 2024-12-18
enh: update validate_api_key to use GET request

jmoralez:
Updates the `validate_api_key` method to perform a `GET /validate_api_key` instead of `POST /validate_token`, since we don't send any data.

Also updates the definition of the method to include a docstring and log the endpoint response.

### DIFF
--- a/nixtla/nixtla_client.py
+++ b/nixtla/nixtla_client.py
@@ -928,19 +928,28 @@ class NixtlaClient:
         return df, X_df, drop_id, freq
 
     def validate_api_key(self, log: bool = True) -> bool:
-        """Returns True if your api_key is valid."""
-        try:
-            with httpx.Client(**self._client_kwargs) as client:
-                validation = self._make_request_with_retries(
-                    client, "validate_token", {}
-                )
-        except:
-            validation = {}
-        if "support" in validation and log:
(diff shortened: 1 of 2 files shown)


nixtla/nixtla_client.py
```diff
@@ -928,19 +928,28 @@ def _run_validations(
...
+            raise NotImplementedError(
+                "validate_api_key is not implemented for Azure deployments, "
+                "you can try using the forecasting methods directly."
+            )
+        with httpx.Client(**self._client_kwargs) as client:
+            resp = client.get("/validate_api_key")
+            body = resp.json()
+        if log:
+            logger.info(body["detail"])
+        return resp.status_code == 200
```
elephaint, line 952: Suppose this returns `False`, how do users know what to do next? It might be informative to suggest an action using a message in case of a `False` return (e.g. that they should go to the dashboard and check their API key?)

elephaint (approved):
LGTM. Non-blocking comment; Personally I think this function would benefit from pointing users in the right direction in case it returns `False` but it's non-blocking for me.

nixtla/nixtla_client.py
jmoralez: That's what `body["detail"]` contains.

PR #576, opened 2024-12-19
enh: use GET request for model_params

jmoralez:
Similar to #573 this changes the call to get the model metadata to a GET request with query parameters.


elephaint (approved):
LGTM

PR #579, opened 2024-12-19
docs: tutorial on reusing fine-tuned models

jmoralez:
Adds a tutorial on how to save, use and delete fine-tuned models.


nbs/docs/tutorials/061_reusing_finetuned_models.ipynb
```diff
@@ -0,0 +1,831 @@
+{
```
ngupta23, line 1: <p>Can we just rename the method to <code>dump</code>? Appending <code>model_</code> seems redundant. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/579/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

ngupta23:
Nice enhancement!

nbs/docs/tutorials/061_reusing_finetuned_models.ipynb
ngupta23, line 1: <p>Do we need to elaborate on what the limit is for storing finetuned models?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/579/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

jmoralez: We haven't settled on the limit, Max wants it to be related to the plan, but right now it's 50 for everyone. I'll set that in the doc for now.

jmoralez: That's a method of pydantic's BaseModel, so I have no control over that. What we could do is something like `NixltaClient.finetuned_models(as_df=True)` and have that return the dataframe instead. WDYT?

ngupta23: <p>Yes, that is a good idea. </p>

jmoralez: added in 35c4ff5

Issue #571, opened 2024-12-18
[docs] Forecasting with multiple time series with missing values


ngupta23:
### Description

The [missing value tutorial](https://docs.nixtla.io/docs/tutorials-missing_values) shows how to handle missing values when only 1 time series is present in the data, but no guidance is given on what to do when there are multiple time series in the data with missing values. In such cases, individual time series can have different start and end dates.

Having a tutorial (or extending the original one) to show how to handle these scenarios would be helpful for users.

Some decisions that the user can take

1. Should the missing time stamps be added from the global min date in the data or on a per series basis (earliest time stamp per series)
2. Should the missing time stamps be added until the global max date in the data or on a per series basis (latest time stamp per series)

The [fill-gaps](https://nixtlaverse.nixtla.io/utilsforecast/preprocessing.html#fill-gaps) utility can be utilized for this purpose but it would be great to highlight the various arguments that the user can use in case of multiple time series.

### Link

https://docs.nixtla.io/docs/tutorials-missing_values

PR #580, opened 2024-12-19
docs: updated to include `fill_gaps` capability for multiple time series

ngupta23:
* Closes https://github.com/Nixtla/nixtla/issues/571

* Updated version of [Missing Values tutorial](https://nixtlaverse.nixtla.io/nixtla/docs/tutorials/missing_values.html) can be seen below.

image


jmoralez:
I'm not a fan of pasting docstrings in a notebook, I'd prefer if you linked to the [function reference](https://nixtlaverse.nixtla.io/utilsforecast/preprocessing.html#fill-gaps) or even just have `fill_gaps?` in a cell.

ngupta23:
> I'm not a fan of pasting docstrings in a notebook, I'd prefer if you linked to the [function reference](https://nixtlaverse.nixtla.io/utilsforecast/preprocessing.html#fill-gaps) or even just have `fill_gaps?` in a cell.

OK, I have updated this using the `show_doc` capability. I think it is a little richer compared to the ? approach.

image

settings.ini
```diff
@@ -16,7 +16,7 @@ custom_sidebar = True
 license = apache2
 status = 4
 requirements = annotated-types httpx[zstd] orjson pandas pydantic>=1.10 tenacity tqdm utilsforecast>=0.2.8
-dev_requirements = black datasetsforecast fire hierarchicalforecast jupyterlab nbdev neuralforecast numpy<2 plotly polars pre-commit pyreadr python-dotenv pyyaml setuptools<70 statsforecast tabulate
+dev_requirements = black datasetsforecast fire hierarchicalforecast jupyterlab nbdev neuralforecast numpy<2 plotly polars pre-commit pyreadr python-dotenv pyyaml setuptools<70 statsforecast tabulate ipywidgets
```
jmoralez, line 19: What is this for?

ngupta23: It is needed to use the `show_docs` capability provided by nbdev.

jmoralez: I don't think so, `uv run --with nbdev --with utilsforecast python -c 'from nbdev import show_doc; from utilsforecast.preprocessing import fill_gaps; print(show_doc(fill_gaps))'` runs fine without it.

ngupta23: It runs but we get this message.

image

jmoralez: that seems to come from nixtla, rather than nbdev. Let's leave it but please note that these are sorted alphabetically, so please set it in the right place.

ngupta23: Done

jmoralez (approved):

PR #591, opened 2025-01-16
[FIX] Contiguity on exogenous

elephaint:
When using `num_partitions != None` and exogenous features are being used, the contiguity of the payload isn't ensured when orjson receives it.


jmoralez (approved):
Thanks!

PR #621, opened 2025-03-19
misc: Fix typing for client timeout

adamantike:
The `timeout` parameter in the `NixtlaClient` class can be set as `None` to disable the timeout, as explained in the docstring.


ngupta23:
@adamantike - Thanks for the PR. We use `nbdev` for developing this repo. You may need to update [the notebook](https://github.com/Nixtla/nixtla/blob/main/nbs/src/nixtla_client.ipynb) instead of the py file and then use `nbdev_export` to create the py files from it. Please see the [instructions here](https://github.com/Nixtla/nixtla/blob/main/CONTRIBUTING.md#pr-submission-guidelines). I hope this helps.

adamantike:
Thanks for the context! I should have checked out the contributing guidelines, but thought it was a one-line fix. PR updated!

jmoralez (approved):

PR #623, opened 2025-03-20
[FEAT] Temporal Hierarchical with TimeGPT

elephaint:
Adds doc page for doing temporal reconciliation with TimeGPT


nbs/docs/tutorials/23_temporalhierarchical.ipynb
```diff
@@ -0,0 +1,758 @@
+{
```
ngupta23, line 1: <blockquote>We are interested in generating forecasts for the hourly and 2-hourly windows. To do so, we make use of hierarchical forecasting techniques.</blockquote><p> </p><p>It is not clear to me why we would need to use hierarchical forecasting to generate the 2 frequency forecast.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/623/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

ngupta23, line 1: <p>Maybe also show what <code>Y_train</code> looks like before we go into this cell, so it is easier to understand.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/623/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

ngupta23, line 1: <p>I wonder if we should have another example here for reconciliation that needs the in-sample predictions for completeness?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/623/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

ngupta23, line 1: <p>Do you want to add the <code>2-hour-period</code> plot for completeness also (with the understanding that y will still be at 1 hour intervals). </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/623/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

ngupta23 (approved):
Looks good to me. Just left some minor comments. Looks like a useful capability, especially when lowest time granularity data is very noisy. Thanks!

nbs/docs/tutorials/23_temporalhierarchical.ipynb
elephaint: Good point, however temporal reconciliation doesn't support reconciliation methods that require in-sample predictions. This is because there are no in-sample predictions for the test set (by definition).

elephaint: <p>Good point, I added an explanation in the cell above</p>

elephaint: <p>I changed the wording</p>

elephaint:
> Looks good to me. Just left some minor comments. Looks like a useful capability, especially when lowest time granularity data is very noisy. Thanks!

Thanks! Indeed, it has a practical use case for improving noisy lower level forecasts. But also in general - the uplift in bottom-level accuracy is higher than any other "improvement technique" (except for adding exogenous, but that adds data) we showcase in our "improving accuracy" notebook. So it's also simply a useful technique for improving forecasting accuracy which can be applied in general.

PR #629, opened 2025-04-09
refactor: Replace httpx.Client with _make_client

Yibei990826:


nixtla/nixtla_client.py
```diff
@@ -978,12 +978,15 @@ def _maybe_override_model(self, model: _Model) -> _Model:
             model = "azureai"
         return model
 
+    def _make_client(self, **kwargs):
```
goodwanghan, line 981: Please add type annotations

PR #644, opened 2025-05-20
[Draft] Remove nbdev from the core development flow

JQGoh:
# Rationale
* Remove the dependency of `nbdev` as part of the development workflow
* Introduce test execution using pytest and provide the test coverage of code base
* Note that we do not remove `nbdev` completely at this stage as it could still be useful to support the generation of notebook-based documentations. 

# Changes 
* Migrate tests from notebook cells to pytest-styled  test cases. The migrated test cases are those mainly registered in.
  - nbs/src/date_features.ipynb
  - nbs/src/nixtla_client.ipynb
  - nbs/src/utils.ipynb

Migrated test cases are saved into `nixtla_tests` folder.
* Move commonly used test fixtures into `nixtla_tests/conftest.py` file
* Revise the ci workflow and contributing guides to omit the dependency on nbdev. 
* As `nbdev_test` also includes execution for notebook in `nbs/docs` folder such as 
`nbs/docs/tutorials/15_missing_values.ipynb` file, we also need to execute and run this. For notebooks saved in the `nbs/docs`, we shall replace with another job like `Run nbconvert`

## Added Packages
 * pytest
 * pytest-cov
 * shap [required as part of the execution of notebook doc]
 * pandas_market_calendars [required part of execution of notebook doc]

### DIFF
--- a/nixtla_tests/conftest.py
+++ b/nixtla_tests/conftest.py
@@ -18,7 +18,7 @@ from utilsforecast.data import generate_series
 from utilsforecast.feature_engineering import fourier, time_features
 from types import SimpleNamespace
 
-load_dotenv()
+load_dotenv(override=True)
 
 
 # note that scope="session" will result in failed test


goodwanghan:
@JQGoh I did some research, and according to [this note](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow) forks can't get the action secrets. So would it be possible to directly create a branch in the Nixtla repo and then a pull request? You should already have permissions to do that.

PR #653, opened 2025-06-17
feat: new date features tutorial

ngupta23:


nbs/docs/tutorials/date_features.ipynb
```diff
@@ -0,0 +1,3321 @@
+{
```
marcopeix, line 1: <p>I would personally remove the "tremendously". I would replace that with a percentage improvement and let the readers make their own judgment.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/nixtla/pull/653/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='54'/>

goodwanghan:
Very nice! I think the file name should start with a number?

PR #677, opened 2025-09-15
[DOCS] add redirect message

deven367:
This PR adds a redirection message in case of a 404. The end result would look like this

image


deven367:
The error was resolved by @eduardofv, hence closing this

PR #678, opened 2025-09-17
Update image paths in documentation to fix broken images

khuyentran1401:


deven367 (approved):
lgtm

loama:
Images seem to be loading correctly in mintlify, but not in github, maybe we need to use the absolute path instead of a relative one?

For example this doc https://github.com/Nixtla/nixtla/blob/main/timegpt-docs/forecasting/exogenous-variables/date_features.mdx shows the markdown for the image instead of the image

PR #691, opened 2025-10-16
DOCS: update support email

loama:


deven367 (changes requested):
As per @JQGoh suggestion, we need to update the `openapi.json` as well

PR #694, opened 2025-10-23
[FEAT]: model_parameters support in cross_validation method 

JQGoh:
# Rationale
* Support `model_parameters` for cross_validation method.
* relax model_parameters check such that we can accept a list of integers etc. too


elephaint (approved):
Nice, one non-blocking comment

PR #816, opened 2026-05-14
[DOC] Update of documentation on the deprecation notice for /v2/historic_forecast

JQGoh:
Attached is the screenshot I generated locally in codespace as preview
image

### DIFF
--- a/timegpt-docs/openapi.json
+++ b/timegpt-docs/openapi.json
@@ -2059,9 +2059,10 @@
     },
     "/v2/historic_forecast": {
       "post": {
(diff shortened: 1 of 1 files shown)


deven367:
@JQGoh just to confirm, have we also updated any other references in the documentation to `v2/historic_forecast` endpoint?

JQGoh:
> @JQGoh just to confirm, have we also updated any other references in the documentation to `v2/historic_forecast` endpoint?

Per my understanding /v2/historic_forecast reference on the API description only involve the openapi.json file

deven367 (approved):
LGTM

PR #823, opened 2026-06-05
chore: target now properly the active route in the sidemenu nav

rslmrn:


rslmrn:
Screenshot 2026-06-04 at 9 03 57 p m

Now active route working with the new mintlify markup for the sidemenu

deven367 (approved):

PR #828, opened 2026-06-09
[FIX] Re-enable skipped tests

JQGoh:
* Fixing some tests to match the latest update of server site logic
~~* Update dependencies like aiohttp and pyarrow to address vulnerability dependencies~~
* delete notebook items related to deprecated weights_x field. Also update the doc descriptions


nixtla_tests/nixtla_client/test_finetune_and_forecast.py
```diff
@@ -120,7 +119,7 @@ def test_get_single_finetuned_model(self, custom_client):
         assert single_model.base_model_id == model_ids_object.model_id1
 
     def test_non_existing_model_returns_error(self, custom_client):
-        with pytest.raises(ApiError, match="Model not found"):
```
JQGoh, line 123: revised error message, expected change

nixtla_tests/nixtla_client/test_client.py
```diff
@@ -186,7 +185,9 @@ def test_setting_one_as_historic_and_other_as_future(
     nixtla_test_client.forecast(
         train, h=5, X_df=future[["unique_id", "ds", "year"]], hist_exog_list=["month"]
     )
-    assert nixtla_test_client.weights_x["features"].tolist() == ["year", "month"]
+
+    # no attribute of weights_x as revised in server-side (no longer has unexpeected weights)
+    assert getattr(nixtla_test_client, "weights_x", None) is None
```
gee-senbong: May be we should use something like
```
with pytest.raises(AttributeError):
    _ = nixtla_test_client.weights_x
```

This assertion is not able to differentiate the following two cases:
1) The attribute does not exist
2) The attribute exists but it is set to `None`

JQGoh: I have just revised the test

gee-senbong (approved):

PR #832, opened 2026-06-29
Update documentation (autogenerated from servers side)

JQGoh:
* Use the newly generated openapi.json from the server


loama (approved):
nice

PR #842, opened 2026-07-18
docs(styles): tokenize style.css into a design system + styling adjustments

rslmrn:
Refactor the Mintlify theme layer into a consistent, token-driven system and
apply a batch of component styling fixes.

Design tokens
- Spacing scale: add --space-6/10/12/14; replace hardcoded px with tokens
- Color: promote the code palette (--blog-code-*) to :root single-source,
  add --code-hover-overlay / --code-tab-hover, dedupe --code-group-border
- Motion: --ease-standard, --duration-fast/base/shift, --transition-hover
- Typography: --font-sans / --font-mono, --font-weight-* scale
- Semantic tints: --nav-tint, --toc-rail; reuse --primary-muted
- Dropdown: --dropdown-hover (light --hover / dark --hover-alt)
- Prune dead tokens; fix --spacing-footer-top typo

Structure & cleanup
- Consistent section banners; trim junk comments
- SYSTEM NOTES block: breakpoints + Mintlify-coupled selectors (upgrade audit)
- Drop unused Supply Ultralight @font-face

Component adjustments
- Header/title max-width, spacing; copy-page CTA responsive swap (viewport)
- Tabs, accordions (grouped, connectors, open border), steps, cards
- Blockquote restyle; code blocks always sharp-cornered (--code-block-radius)
- Grain layer extended to theme menu, mobile menu, search button
- Prose heading/list/hr spacing; table + dropdown theming

### DIFF
--- a/timegpt-docs/accordion-scroll-lock.js
+++ b/timegpt-docs/accordion-scroll-lock.js
@@ -10,7 +10,10 @@
  */
 (function () {
   var SEL = "details.accordion > summary, details.expandable > summary";
-  var PIN_MS = 400; // covers the 0.4s open animation + smooth-scroll frames
+  // Covers the open animation + smooth-scroll frames. Coupled by value to the
+  // CSS token --duration-slow (0.4s) that drives the accordion animation — if
+  // that token changes, keep this in sync (no shared source across CSS/JS).
+  var PIN_MS = 400;
 
   document.addEventListener(
     "click",
--- a/timegpt-docs/accordion-scroll-lock.js
+++ b/timegpt-docs/accordion-scroll-lock.js
@@ -10,7 +10,7 @@
  */
 (function () {
   var SEL = "details.accordion > summary, details.expandable > summary";
-  var PIN_MS = 700; // covers the 0.4s open animation + smooth-scroll frames
+  var PIN_MS = 400; // covers the 0.4s open animation + smooth-scroll frames
 
   document.addEventListener(
     "click",
@@ -29,6 +29,6 @@
       }
       requestAnimationFrame(pin);
     },
-    true
+    true,
   );
 })();
--- a/timegpt-docs/accordion-scroll-lock.js
+++ b/timegpt-docs/accordion-scroll-lock.js
@@ -1,0 +1,34 @@
+/*
+ * Accordion open/close: keep the scroll position steady.
+ *
+ * Mintlify deep-links accordions — opening one sets location.hash to the
+ * accordion id, which makes the browser scroll to that anchor (scroll-margin-top
+ * + html { scroll-behavior: smooth }). That reads as the page "jumping". CSS
+ * can't cancel fragment navigation, so we pin the scroll position for a short
+ * window after a summary click, defeating the anchor scroll while KEEPING the
+ * hash (the shareable link still works). Runs in dev and prod.
+ */
+(function () {
+  var SEL = "details.accordion > summary, details.expandable > summary";
+  var PIN_MS = 700; // covers the 0.4s open animation + smooth-scroll frames
+
+  document.addEventListener(
+    "click",
+    function (e) {
+      var summary = e.target.closest && e.target.closest(SEL);
+      if (!summary) return;
+
+      var y = window.scrollY;
+      var start = -1;
+
+      function pin(now) {
+        if (start < 0) start = now;
+        // Only correct if something tried to move us — avoids fighting the user.
+        if (window.scrollY !== y) window.scrollTo(0, y);
+        if (now - start < PIN_MS) requestAnimationFrame(pin);
+      }
+      requestAnimationFrame(pin);
+    },
+    true
+  );
+})();
--- a/timegpt-docs/code-group-tab-labels.js
+++ b/timegpt-docs/code-group-tab-labels.js
@@ -8,7 +8,7 @@
  * no title. Re-runs on client-side nav and when accordions reveal code groups.
  */
 (function () {
-  var FALLBACK = "Example code";
+  var FALLBACK = "Example";
 
   function labelTab(tab) {
     var titleBox = tab.querySelector('[class*="peer/title"]');
@@ -18,7 +18,7 @@
     var fn =
       panel &&
       panel.querySelector(
-        '[data-component-part="code-block-header-filename"] span'
+        '[data-component-part="code-block-header-filename"] span',
       );
     var text = (fn && fn.textContent.trim()) || FALLBACK;
 
@@ -32,9 +32,7 @@
   }
 
   function run() {
-    document
-      .querySelectorAll('.code-group [role="tab"]')
-      .forEach(labelTab);
+    document.querySelectorAll('.code-group [role="tab"]').forEach(labelTab);
   }
 
   var scheduled = false;
--- a/timegpt-docs/code-group-tab-labels.js
+++ b/timegpt-docs/code-group-tab-labels.js
@@ -48,8 +48,24 @@
   if (document.readyState !== "loading") run();
   else document.addEventListener("DOMContentLoaded", run);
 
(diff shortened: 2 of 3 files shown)


timegpt-docs/style.css
```diff
@@ -402,143 +922,687 @@ a.link {
...
+}
+
+.prose
+  :where(blockquote):not(:where([class~="not-prose"], [class~="not-prose"] *))
+  :where(p)
+  + :where(p) {
+  margin-top: var(--space-16) !important;
+}
+
+[data-component-part="card-content-container"] {
```
nasaul: **Dead / conflicting duplicate rule (cleanup).** `[data-component-part="card-content-container"]` is defined here with `padding: 0` and again at ~line 1710 with `padding: var(--space-32) var(--space-32)`. Same selector, same specificity, both `!important` — the later rule's `padding` shorthand wins, so all three padding declarations in *this* block are dead code. Remove this block (or scope it to the context it was actually meant for) to avoid confusing future readers.

timegpt-docs/style.css
```diff
@@ -571,146 +1657,246 @@ details.accordion :not(summary) span[data-as="p"] {
...
-
 html.dark a.card svg,
 html.dark div.card svg {
   background: var(--primary);
 }
 
 /* Content container */
 [data-component-part="card-content-container"] {
-  padding: 40px 30px !important;
+  padding: var(--space-32) var(--space-32) !important;
```
nasaul: Minor: `var(--space-32) var(--space-32)` is redundant — a single `var(--space-32)` is equivalent. (See also the conflicting duplicate definition of this selector at ~line 1458.)

timegpt-docs/style.css
```diff
@@ -67,35 +66,106 @@
...
+
+  /* FONTS */
+  --font-sans: "NeueMontreal", sans-serif;
+  --font-mono: "Supply", monospace;
+
+  /* FONTS WEIGHT */
+  --font-weight-regular: 400;
+  /* Override medium from mintlify to our own 400 weight instead of changing it in mintlify classes, our DLS only uses 400 weight */
+  --font-weight-medium: 400;
+  --font-weight-semibold: 400;
```
nasaul, line 113: **Confirm this is intended.** With `--font-weight-semibold: 400`, `.prose strong` and table `<th>` render at 400 — so `**bold**` markdown and header cells become visually indistinguishable from body text across all docs prose. If that matches the DLS, fine, but it's a real emphasis/readability change, not just token plumbing. Please confirm it's deliberate.

timegpt-docs/style.css
```diff
@@ -1020,19 +2313,109 @@ body::after {
   content: "";
   position: fixed;
   inset: 0;
-  z-index: 9999;
+  z-index: 10;
```
nasaul, line 2310: **Grain z-index 9999 → 10 needs visual QA.** The `body::after` grain dropped from 9999 to 10 (sidebar to 11) and grain is now also layered onto navbar/banner/sidebar/menus/dialogs via per-element `::after`. Sensible move (grain is `pointer-events: none`), but it's a meaningful stacking change — worth a deliberate pass over modals, mobile menu, search, and the chat sheet in both themes to make sure grain neither disappears behind a newly-stacked surface nor doubles up.

timegpt-docs/style.css
```diff
@@ -571,146 +1657,246 @@ details.accordion :not(summary) span[data-as="p"] {
...
+details.accordion {
+  border-color: var(--border) !important;
+}
+
+/* Open/close animation. ::details-content is the generated wrapper around the
+   non-summary content; interpolate-size lets its height:auto animate. overflow
+   hidden clips during the slide and prevents the content margins collapsing out.
+   Degrades to an instant toggle where ::details-content isn't supported. */
+details.accordion {
+  interpolate-size: allow-keywords;
```
nasaul, line 1800: `::details-content` + `interpolate-size: allow-keywords` are Chromium-only in this era; Safari/Firefox fall back to an instant toggle. The comment already calls this out (good) — just confirm the instant-toggle fallback and the surrounding `:has()`-heavy rules render correctly in Safari/Firefox, since the docs audience is broad.

nasaul (commented):
Thorough, well-documented theme refactor — the `SYSTEM NOTES` audit block and the graceful-degradation comments are great practice. Posting the review as inline comments. Nothing here is a hard correctness blocker in the JS; the main things to resolve before merge are the dead duplicate rule, confirming the `semibold: 400` bold-flattening is intentional, and verifying the two root `.js` files actually load in the built site. The rest are QA-and-monitor items.

timegpt-docs/accordion-scroll-lock.js
```diff
@@ -0,0 +1,34 @@
...
+ * Mintlify deep-links accordions — opening one sets location.hash to the
+ * accordion id, which makes the browser scroll to that anchor (scroll-margin-top
+ * + html { scroll-behavior: smooth }). That reads as the page "jumping". CSS
+ * can't cancel fragment navigation, so we pin the scroll position for a short
+ * window after a summary click, defeating the anchor scroll while KEEPING the
+ * hash (the shareable link still works). Runs in dev and prod.
+ */
+(function () {
+  var SEL = "details.accordion > summary, details.expandable > summary";
+  var PIN_MS = 400; // covers the 0.4s open animation + smooth-scroll frames
```
nasaul: Minor coupling: `PIN_MS = 400` duplicates `--duration-slow: 0.4s` by value — if the token changes, this won't follow. A comment noting the coupling (or a shared source) would prevent future drift.

Separately: neither this file nor `code-group-tab-labels.js` is referenced in `docs.json`, so both rely on Mintlify's root-asset auto-load convention. Please verify they're actually injected in the built/preview site (network tab or page source) — if not wired up, both features silently no-op with no error.

timegpt-docs/code-group-tab-labels.js
```diff
@@ -0,0 +1,56 @@
...
+      scheduled = false;
+      run();
+    });
+  }
+
+  if (document.readyState !== "loading") run();
+  else document.addEventListener("DOMContentLoaded", run);
+
+  // Re-run when tabs/panels mount late (hydration, client-side nav, accordions).
+  new MutationObserver(schedule).observe(document.body, {
```
nasaul: Minor perf: this observes `document.body` with `childList + subtree`, so every DOM mutation anywhere schedules a full-document `querySelectorAll('.code-group [role="tab"]')`. It's rAF-debounced and converges (the `label.textContent !== text` guard stops the self-triggered loop), so it's acceptable — but on long API/reference pages it's a bit heavy. If you see jank, narrow the observed subtree or disconnect once tabs are stable.

timegpt-docs/style.css
rslmrn: Confirmed. First, a note: the z-index: 9999 is already gone — the rule now has a single z-index: 10 (style.css:2310), so the duplicate declaration is resolved.

timegpt-docs/style.css
rslmrn: Confirmed — the fallback is safe by construction. ::details-content is an all-or-nothing pseudo-element: a browser either supports it (Chromium → applies height:0 and animates it) or it doesn't (Safari/Firefox → the entire rule is invalid and dropped). So on Safari/Firefox:
- height: 0 / overflow: hidden are never applied — there's no "content stuck collapsed at height 0" failure mode, because the collapse and the animation live in the same dropped rule.
- The native <details>/[open] toggle takes over → instant open/close, exactly the documented degradation.
 - interpolate-size: allow-keywords unsupported → ignored, harmless.

timegpt-docs/style.css
rslmrn: Its intended in order to override the mintlify variables instead of updating a lot of components with this rule. Our DLS only uses 400 as font-weight

nasaul (approved):
LGTM

PR #852, opened 2026-08-04
[FEAT] Simulate and explain

elephaint:
Adds endpoints:

- simulate — generate temporally correlated sample paths (Schaake shuffle), with optional cross-series coupling (multivariate=True), custom quantile grids, seeds, and partitioned requests.
- explain — model-independent historical feature importance via Granger causality or transfer entropy.
- forecast — new feature_contributions_type parameter (shapley/intervention/granger/transfer_entropy).
- Fixes a silent future-exogenous misalignment bug; X_df is now strictly validated (exactly one row per future (id, time) pair — previously oversized frames were accepted).
- Docs, tutorials, and notebooks for both endpoints; extensive offline test suites.
- Duplicate timestamps are now rejected across all endpoints (behavioral change)

Async job behavior:

- All endpoints share polling and cancellation. Job submissions, including `execute_step`, retry only connection failures and HTTP 429; read timeouts and gateway errors fail immediately to avoid duplicate jobs. Submission retry sleeps respect the remaining `max_wait_time` budget.
- Async partitioned requests run at most five jobs concurrently. A partition failure cancels sibling jobs and stops queued submissions; synchronous partitioned requests also stop queued work after failure.
- `simulate` and `explain` use the existing `AsyncJobTimeoutError` and `AsyncJobCancelledError` classes. `AsyncJobError.error` preserves the original server error. Permanent polling errors remain intact, and automatic submit-and-wait calls request cleanup when the terminal state is unknown.
- `Job.wait()` requires finite polling settings; the new endpoints continue to support `async_job_wait_timeout=None`. Cancellation races, retry budgets, cleanup and exception behavior have offline regression coverage.


marcopeix (commented):
If you look at the preview, some icons do not render in Mintlify, not sure why. Other than that, the new documentation looks really good!

nasaul (commented):
Two things I'd like called out before merge, and five things I think we can delete.

## Breaking changes not in the PR description

**1. Duplicate timestamps are now rejected on every endpoint.** `_validate_freq_regularity` adds a `_has_duplicate_keys` check, and it's called from `forecast`, `cross_validation`, `detect_anomalies`, and `detect_anomalies_online` — not just the new endpoints. Data with duplicate `(id_col, time_col)` rows that happened to match the expected grid length used to pass and now raises `ValueError`.

**2. `X_df` is now strictly validated.** An oversized `X_df` used to be accepted (silently misaligned); it now raises.

Both are correct fixes and both should land. But they're behavior changes for existing users on a PR titled `[FEAT]`, so they belong in the description and the release notes rather than being discovered in production. Worth deciding whether they ship here or split into their own PR.

## Suggested cuts

**1. Two async-job stacks now coexist.** `_run_async_job` + `_make_partitioned_requests(_is_async_job=True)` (forecast/finetune/cross_validation) now sits alongside `_run_async_task` + `_poll_async_job` + `_dispatch_async_jobs` (simulate/explain). The new path is strictly better — the old one still carries the `NOTE:` comment admitting that siblings keep running after one partition fails, which `_dispatch_async_jobs` solves via the shared `cancellation_event`. Migrating forecast onto the new path is a net deletion and fixes that NOTE for free. Happy for this to be a follow-up PR rather than a blocker, but I'd rather not leave two dispatchers in place for long.

**2. Two literals that guess at our own server's behavior.**

```python
_ASYNC_JOBS_UNAVAILABLE_MARKER = "not available in this deployment"
_ASYNC_JOB_ID_PREFIXES: dict[str, str] = {"simulate": "sm", "explain": "ex"}
```

The first substring-matches the server's error prose — reword that message and the branch dies silently, with no test failure. The second hard-fails a job that ran fine if the ID scheme ever changes. A 503 already carries the server's own explanation; I'd let it through unmodified and drop the job-id prefix assertion entirely.

**3. Echo-back validation.** `simulate` re-checks that the response's `n_paths`, `h`, and `coupled` match what it just sent, and `explain` re-checks `method`. `_coerce_coupled_flag` is six lines asserting that a bool is a bool. The `samples.size != output_values` and `response_sizes` checks are worth keeping — those catch real corruption before it becomes a wrong forecast — but the metadata echoes only catch a server bug we'd hear about another way.

**4. Four helpers that are one line each.** `_sleep` and `_wait_for_poll` are two module-level functions for `event.wait(s) if event else time.sleep(s)`. `_sleep` appears to exist only so `test_simulate_explain_async_jobs.py` can monkeypatch it, and `monkeypatch.setattr(time, "sleep", ...)` already covers that. `_sort_categorical_values` is a dict comprehension and `_log_exog_features` is four `logger.info` calls; both read fine inlined at their two call sites.

**5. Seed derivation.**

```python
seed_span = _MAX_SEED - _MIN_SEED + 1
part["seed"] = (seed - _MIN_SEED + i) % seed_span + _MIN_SEED
```

`-(2**63)` through `2**64 - 1` spans no actual integer type, and the constants' own comment notes the range is validated server-side. `part["seed"] = seed + i` gives the same distinctness in one line and lets the server reject anything out of range.

## Smaller notes

- `explain(method=...)` is typing-only, so a typo costs a full async submit and comes back as an opaque `AsyncJobError`. `simulate` validates `h`, `n_paths`, `num_partitions`, and `seed` up front — worth a matching one-liner.
- `simulate` returns `coupled` as a full column: `np.full(output_values, coupled)` writes the same bool to every one of `n_paths × n_series × h` rows. It's request metadata, not per-row data.
- `_run_async_task` starts the `async_job_wait_timeout` clock after submission, so submit retries stack on top of it and worst-case wall clock is the sum of the two. Fine as behavior, but worth a line in the docstring since users will size the timeout against total wait.
- `_collect_concurrent_results` is shared by two of the three dispatchers; `_dispatch_async_jobs` reimplements the loop because it needs cancellation. Resolves itself if we do cut 1.

nixtla/nixtla_client.py
```diff
@@ -1249,18 +1764,299 @@ def _get_request(
...
+        prefix = f"{_ASYNC_JOB_ID_PREFIXES[task]}-"
+        if not isinstance(job_id, str) or not job_id.startswith(prefix):
+            raise RuntimeError(
+                f"Unexpected response when submitting the {task} job: {body}"
             )
-        if resp.status_code != 200:
-            raise ApiError(status_code=resp.status_code, body=resp_body)
-        return resp_body
+        logger.info(f"Submitted {task} job {job_id}. Waiting for it to finish...")
+        return job_id
```
gee-senbong: Any reason not to reuse existing `_submit_job`,  `_run_async_job`  & `_poll_job`?

nixtla/nixtla_client.py
```diff
@@ -331,6 +485,127 @@ def _maybe_infer_freq(
...
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=pd.errors.PerformanceWarning)
+            expected_ends = times_by_id["min"] + freq * (times_by_id["size"] - 1)
+        freq_ok = bool((expected_ends == times_by_id["max"]).all())
+    else:
+        raise ValueError(
+            "`freq` should be a string, integer or pandas offset, "
+            f"got {type(freq).__name__}."
+        )
+    if freq_ok and _has_duplicate_keys(df, id_col, time_col):
```
JQGoh: nit: `_has_duplicate_keys` by right can conclude whether we have a duplicate-key issue. So we could consider suggesting a more specific error message like duplicate timestamp instead of "Series contain missing or duplicate timestamps, or the timestamps do not match the provided frequency." As this message covers a few scenarios that can raise error; if we can be specific, we can guide the user to inspect the most relevant raised error.

elephaint:
> Two things I'd like called out before merge, and five things I think we can delete.
> ## Breaking changes not in the PR description
> 
> **1. Duplicate timestamps are now rejected on every endpoint.** `_validate_freq_regularity` adds a `_has_duplicate_keys` check, and it's called from `forecast`, `cross_validation`, `detect_anomalies`, and `detect_anomalies_online` — not just the new endpoints. Data with duplicate `(id_col, time_col)` rows that happened to match the expected grid length used to pass and now raises `ValueError`.
> 
> **2. `X_df` is now strictly validated.** An oversized `X_df` used to be accepted (silently misaligned); it now raises.
> 
> Both are correct fixes and both should land. But they're behavior changes for existing users on a PR titled `[FEAT]`, so they belong in the description and the release notes rather than being discovered in production. Worth deciding whether they ship here or split into their own PR.
> ## Suggested cuts
> 
> **1. Two async-job stacks now coexist.** `_run_async_job` + `_make_partitioned_requests(_is_async_job=True)` (forecast/finetune/cross_validation) now sits alongside `_run_async_task` + `_poll_async_job` + `_dispatch_async_jobs` (simulate/explain). The new path is strictly better — the old one still carries the `NOTE:` comment admitting that siblings keep running after one partition fails, which `_dispatch_async_jobs` solves via the shared `cancellation_event`. Migrating forecast onto the new path is a net deletion and fixes that NOTE for free. Happy for this to be a follow-up PR rather than a blocker, but I'd rather not leave two dispatchers in place for long.
> 
> **2. Two literals that guess at our own server's behavior.**
> 
> ```python
> _ASYNC_JOBS_UNAVAILABLE_MARKER = "not available in this deployment"
> _ASYNC_JOB_ID_PREFIXES: dict[str, str] = {"simulate": "sm", "explain": "ex"}
> ```
> 
> The first substring-matches the server's error prose — reword that message and the branch dies silently, with no test failure. The second hard-fails a job that ran fine if the ID scheme ever changes. A 503 already carries the server's own explanation; I'd let it through unmodified and drop the job-id prefix assertion entirely.
> 
> **3. Echo-back validation.** `simulate` re-checks that the response's `n_paths`, `h`, and `coupled` match what it just sent, and `explain` re-checks `method`. `_coerce_coupled_flag` is six lines asserting that a bool is a bool. The `samples.size != output_values` and `response_sizes` checks are worth keeping — those catch real corruption before it becomes a wrong forecast — but the metadata echoes only catch a server bug we'd hear about another way.
> 
> **4. Four helpers that are one line each.** `_sleep` and `_wait_for_poll` are two module-level functions for `event.wait(s) if event else time.sleep(s)`. `_sleep` appears to exist only so `test_simulate_explain_async_jobs.py` can monkeypatch it, and `monkeypatch.setattr(time, "sleep", ...)` already covers that. `_sort_categorical_values` is a dict comprehension and `_log_exog_features` is four `logger.info` calls; both read fine inlined at their two call sites.
> 
> **5. Seed derivation.**
> 
> ```python
> seed_span = _MAX_SEED - _MIN_SEED + 1
> part["seed"] = (seed - _MIN_SEED + i) % seed_span + _MIN_SEED
> ```
> 
> `-(2**63)` through `2**64 - 1` spans no actual integer type, and the constants' own comment notes the range is validated server-side. `part["seed"] = seed + i` gives the same distinctness in one line and lets the server reject anything out of range.
> ## Smaller notes
> 
>     * `explain(method=...)` is typing-only, so a typo costs a full async submit and comes back as an opaque `AsyncJobError`. `simulate` validates `h`, `n_paths`, `num_partitions`, and `seed` up front — worth a matching one-liner.
> 
>     * `simulate` returns `coupled` as a full column: `np.full(output_values, coupled)` writes the same bool to every one of `n_paths × n_series × h` rows. It's request metadata, not per-row data.
> 
>     * `_run_async_task` starts the `async_job_wait_timeout` clock after submission, so submit retries stack on top of it and worst-case wall clock is the sum of the two. Fine as behavior, but worth a line in the docstring since users will size the timeout against total wait.
> 
>     * `_collect_concurrent_results` is shared by two of the three dispatchers; `_dispatch_async_jobs` reimplements the loop because it needs cancellation. Resolves itself if we do cut 1.

- Addressed: Removed special 503 message matching and job-ID prefix checks; added upfront explain(method=...) validation; clarified timeout semantics.
- Already covered: Breaking-change documentation is in the PR description.
- ~~Deferred: Async dispatcher consolidation, shared result collection, and sibling cancellation.~~ Edit: also addressed
- Kept unchanged: Response validation, helper functions, seed wrapping, and the coupled column—they preserve useful behavior and correctness.

nixtla/nixtla_client.py
elephaint: Fixed

nixtla/nixtla_client.py
elephaint: The new helpers are a bit more generic, leaving as a follow-up unification. But on second thought, I think I should extend the existing helpers.

elephaint: Fixed that

nasaul (approved):
Everything from my earlier review is addressed — the two async-job stacks are unified, the guessy literals are gone, `explain(method=...)` validates up front, and the duplicate-timestamp error message is now specific. Ran the affected offline test suites against this branch: 306 passed (3 errors are live-endpoint tests needing an API key, unrelated).

One remaining nit, not a blocker: the duplicate-timestamp behavior change (now rejected on `forecast`/`cross_validation`/`detect_anomalies`/`detect_anomalies_online`, not just the new endpoints) still isn't called out in the PR description or CHANGELOG, unlike the `X_df` breaking change which is. Worth a one-line addition before release notes go out.

PR #859, opened 2026-08-14
[FEAT] Start datetime information

JQGoh:
* Registeration of series starting datetime info

### DIFF
--- a/nixtla_tests/nixtla_client/test_nixtla_client.py
+++ b/nixtla_tests/nixtla_client/test_nixtla_client.py
@@ -129,7 +129,7 @@ def test_cv_forecast_consistency(nixtla_test_client, cv_series_with_features):
             X_df=valid,
         )
         np.testing.assert_allclose(
-            cv_res["TimeGPT"], fcst_res["TimeGPT"], atol=1e-4, rtol=1e-3
+            cv_res["TimeGPT"], fcst_res["TimeGPT"], atol=1e-4, rtol=5e-3
         )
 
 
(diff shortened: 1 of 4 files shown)


nixtla_tests/nixtla_client/test_nixtla_client.py
```diff
@@ -129,7 +129,7 @@ def test_cv_forecast_consistency(nixtla_test_client, cv_series_with_features):
             X_df=valid,
         )
         np.testing.assert_allclose(
-            cv_res["TimeGPT"], fcst_res["TimeGPT"], atol=1e-4, rtol=1e-3
+            cv_res["TimeGPT"], fcst_res["TimeGPT"], atol=1e-4, rtol=5e-3
```
marcopeix: why?

marcopeix (commented):
Can we also add a polars test? Just to make sure that timestamps in both polars and pandas work.

elephaint (commented):
@JQGoh I agree with @marcopeix comment. If that is a timegpt-2.1 issue, we should investigate rather than raising the tolerance. I reverted that change and CI seems to pass without issue.

I tried to solve some of the other issues I found:
- Polars needs testing
- Polars/pandas discrepancies
- Ignoring timezones with non-fixed-offsets

The later might pose a serious issue - we're now simply ignoring those, but I didn't investigate how much would be ignored. Other options to fix this would be:
- Ignore the non-fixed offset and have a lossy reconstruction with potential misalignment
- Just send the full ds column

Wdyt?

JQGoh:
@elephaint 
Thanks for pointing out the non-fixed-offsets.

I shall go with registering the first start_datetime instead of passing the whole series time stamps. For non-fixed offset for timezone aware data, it is fine to ignore the registration of these specific cases.
Note that in https://github.com/Nixtla/nixtla/pull/859/changes/2cb4ff378aef65bdc0a3d37ebd8d1e0487ffc1ce, I revised it such that it can still properly register the start_datetime for those timezones with fixed offset.

elephaint (approved):
Looks good to me, thanks 👍

PR #860, opened 2026-08-17
docs(api): document simulate and explain endpoints

loama:
## What

Adds the `/v2/simulate` and `/v2/explain` operations to the public OpenAPI contract, including their complete request and response schemas.

## Why

The routes already exist in `nixtla-engine`, but the OpenAPI file used by Mintlify omits them. Raw API users therefore cannot discover either endpoint in the generated API reference.

## How

1. Synced both authenticated operations from the current `nixtla-engine` OpenAPI generator.
2. Synced every new transitive component schema from the TSFM commit pinned by `nixtla-engine`.
3. Limited the sync to the two new operations and their new schemas so existing documentation corrections remain unchanged.
4. Kept the existing Mintlify configuration because it already generates reference pages from this OpenAPI file.

## Testing

1. [x] Ran the same OpenAPI generator command used by the `nixtla-engine` release workflow in an isolated checkout.
2. [x] Compared both operations and every new component schema structurally against the generated output.
3. [x] Removed the new entries and verified that the remainder is structurally identical to `nixtla` main.
4. [x] Parsed the updated document with `jq` and verified that every component reference resolves.
5. [x] Verified both generated API reference pages in the Mintlify preview, including route paths, request fields, and response fields.
6. [x] Ran `git diff --check` and scanned the complete diff for secrets and unexpected files.

## Checklist

1. [x] Title follows conventional commits format.
2. [x] PR is focused on one concern.
3. [x] Complete diff reviewed.
4. [x] Documentation updated.
5. [x] No secrets, credentials, personal data, generated binaries, or unrelated files included.


elephaint:
Unnecessary as OpenAPI docs are synced automatically @JQGoh

PR #864, opened 2026-08-21
feat: add execute_step support to NixtlaClient

gee-senbong:
Adds `execute_step` to `NixtlaClient`: run one TSMP top-level API call server-side as an async job.

## New

- **`NixtlaClient.submit_execute_step_job(func_name, params, data=None, job_timeout_seconds=None)`** — submits the step and returns a `Job`; `job.wait()` returns a `StepResult`.
- **`nixtla.StepResult`** — mapping of table name to `pyarrow.Table`, plus `.metadata`. Pass `.data` straight into the next step to chain calls.
- **`nixtla.ref(key)`** — builds the `{"data_ref": key}` envelope `params` uses to reference a table in `data`.
- **`nixtla/steps.py`** — the client-side codec. Validates locally (references, table count, body size, metadata size and nesting) so a malformed request fails at the call site instead of surfacing later as a failed job.

## Behaviour changes

- **`Job.wait(cancel_on_timeout=...)` now defaults to `True`.** Previously a `poll_timeout` left the job running server-side. Code that polls in short increments and calls `wait()` again to resume must now pass `cancel_on_timeout=False`.
- A binary job's result fetch polls in its own bounded loop rather than going through the failure-retry path. Waiting no longer logs at ERROR, and giving up raises `AsyncJobTimeoutError` instead of a bare `ApiError` carrying a 202.
- `forecast()` / `cross_validation()` accept `_job_timeout_seconds` when running their work as async jobs. The per-job timeout was previously reachable only through the `submit_*_job` methods.

## Notes

- `pyarrow` moves from a dev dependency to a runtime one (`>=14,<24`). This endpoint exchanges arrow tables, and a pandas round-trip drops the schema metadata that makes chaining lossless.
- A pandas index is never sent as data: an unnamed index is dropped, and a named one raises so the caller decides whether those values belong in a column.

## Testing

`nixtla_tests/nixtla_client/test_execute_step_job.py` (new) plus additions to `test_async_jobs.py`. Fully mocked, no network — 137 offline tests pass.

### DIFF
--- a/nixtla/nixtla_client.py
+++ b/nixtla/nixtla_client.py
@@ -246,9 +246,9 @@ def _is_retriable_error(exc: Exception) -> bool:
     retriable_codes = [
         HTTPStatus.REQUEST_TIMEOUT,
         # A binary job's result endpoint answers 202 while the zip has not materialized yet, so
-        # "not ready" means poll again rather than give up. 409 stays listed because older servers
-        # used it for that same case; on current ones it means the job ended without a result, and
-        # retrying is merely wasteful rather than wrong.
+        # "not ready" means poll again rather than give up. 409 is listed alongside it for
+        # compatibility; where it instead means the job ended without a result, retrying is
+        # merely wasteful rather than wrong.
         HTTPStatus.ACCEPTED,
         HTTPStatus.CONFLICT,
         HTTPStatus.TOO_MANY_REQUESTS,
@@ -1407,9 +1407,9 @@ class NixtlaClient:
         """Fetch a binary job's result from its dedicated endpoint.
 
         Binary results cannot be inlined into the JSON status response, so they are served
-        separately. The server answers 202 while the job has not yet succeeded (older servers
-        answered 409), and `_is_retriable_error` treats both as retriable -- callers should go
-        through `_retry_strategy` so that race resolves itself rather than surfacing.
+        separately. The API answers 202 while the job has not yet succeeded, and 409 is treated
+        the same way; `_is_retriable_error` covers both -- callers should go through
+        `_retry_strategy` so that race resolves itself rather than surfacing.
 
         The status check stays exact: anything other than 200 raises rather than returning a
         body, so a "not ready" JSON payload is never mistaken for the zip.
@@ -4055,9 +4055,9 @@ class NixtlaClient:
     ) -> Job:
         """Submit a single TSMP step to run asynchronously.
 
-        `execute_step` runs one TSMP top-level API call server-side in a
-        fresh sandbox: nothing is registered and nothing is cached between
-        calls, so every request is self-contained and carries its own data.
+        `execute_step` runs one TSMP top-level API call server-side. Each
+        call is independent: no state carries over from one to the next, so
+        every request is self-contained and carries its own data.
         This does not block; it submits the job and immediately returns a
         `Job` handle. Call `job.wait()` to poll until it completes and get a
         `StepResult`, or `job.cancel()` to request that the server stop it.
@@ -4086,11 +4086,11 @@ class NixtlaClient:
         `step.to_pandas()`.
 
         There is no `model` argument: a step names the models it runs inside
-        `params`, and the sandbox reaches them over the API rather than from
-        a checkpoint local to one host.
+        `params`.
 
-        Not reachable over this transport: `train` and `optimize_model`.
-        Both require a `tune.Space`, which has no JSON encoding.
+        Not reachable over this transport: `optimize_model`, which requires a
+        `tune.Space` that has no JSON encoding. A pandas index is never sent
+        as data; a named one is folded into a column, anything else dropped.
 
         Args:
             func_name (str): TSMP top-level API to run, e.g. `'forecast'`,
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "narwhals>=2.11.0",
     "orjson",
     "pandas<3.0.0",
+    "pyarrow>=14.0.0,<24.0.0",
```
JQGoh, line 32: could you share why `pyarrow>=14.0.0`? This question is more out of curiousity and not a blocker to this PR

nixtla/nixtla_client.py
```diff
@@ -246,6 +262,31 @@ def _is_retriable_error(exc: Exception) -> bool:
...
 
+def _validate_job_timeout_seconds(job_timeout_seconds: Optional[int]) -> None:
+    """Reject a job timeout the server would refuse, before spending a round-trip on it.
+
+    Kept identical in wording to the check `steps.build_request` runs for `execute_step`, which
+    validates separately because that module is a self-contained codec.
+    """
+    if job_timeout_seconds is not None and job_timeout_seconds <= 0:
+        raise ValueError(
+            f"job_timeout_seconds must be positive, got {job_timeout_seconds!r}"
```
JQGoh, line 273: No need to enforce a reasonable upper bound like acceptable job_timeout_seconds?

nixtla/nixtla_client.py
```diff
@@ -1300,18 +1358,173 @@ def _cancel_job(self, client: httpx.Client, job_id: str) -> None:
...
         with self._make_client(**self._client_kwargs) as client:
             job_id = self._submit_job(client, endpoint, payload)
-        return Job(client=self, job_id=job_id, endpoint=endpoint, parse_result=parse_result)
+        return Job(
+            client=self,
+            job_id=job_id,
+            endpoint=endpoint,
+            # A JSON result is inline in the status response, so there is nothing to wait for
+            # and the poll settings are unused.
+            get_result=lambda job_data, poll_interval, poll_timeout: parse_result(
```
JQGoh: `poll_interval`, `poll_timeout` are not defined (dummy), could we use '_' to replace them?

nixtla/nixtla_client.py
```diff
@@ -1300,18 +1358,173 @@ def _cancel_job(self, client: httpx.Client, job_id: str) -> None:
...
+        self,
+        client: httpx.Client,
+        endpoint: str,
+        job_id: str,
+    ) -> tuple[httpx.Headers, bytes]:
+        """Fetch a binary job's result from its dedicated endpoint, once.
+
+        Binary results cannot be inlined into the JSON status response, so they are served
+        separately. Callers go through `_wait_for_job_result_bytes`, which handles the
+        not-ready answer this can raise.
```
JQGoh: The saying "which handles the not-ready answer this can raise." does not explain the intention in a clear manner, consider rephrasing this doc string

nixtla/nixtla_client.py
```diff
@@ -3873,6 +4157,95 @@ def submit_cross_validation_job(
...
+        `StepResult`, or `job.cancel()` to request that the server stop it.
+
+        Tables are referenced from `params` with `nixtla.ref(key)`, naming a
+        key of `data`. Because a step's output tables can be passed straight
+        back in as the next step's `data`, calls chain without any file or
+        byte handling::
+
+            from nixtla import ref
+
+            step1 = nc.submit_execute_step_job(
```
JQGoh: `nc` is the nixtla-client? We should explicitly define this for clarity

JQGoh (approved):
Left some non-blocking comments

pyproject.toml
gee-senbong: It's a wheel-availability floor rather than an API one. 14.0.0 is the first pyarrow release that ships `cp312` wheels (13.0.0 ships only cp38-cp311), and we declare `requires-python = ">=3.10"` with CI on 3.10-3.13 — below 14, a Python 3.12 install finds no wheel and falls back to building Arrow C++ from source. Nothing in `steps.py` needs a recent API; it only uses `pa.Table`, `pq.write_table` and `pq.read_table`. The `<24.0.0` ceiling carries over unchanged from the dev dependency on `main`.

One thing worth a follow-up though: 14.0.0 is the *last* version affected by CVE-2023-47248 / GHSA-5wvp-7f3h-6wmm (critical — arbitrary code execution via deserialization in the IPC and Parquet readers, fixed in 14.0.1), and `steps._unpack` calls `pq.read_table` on the server response body, so that is exactly the affected code path. Bumping the floor to `>=14.0.1` would exclude the one vulnerable version at no practical cost. Keeping it out of this PR so the lockfile refresh stays separate.

nixtla/nixtla_client.py
gee-senbong: Deliberately not — the server owns that ceiling and already enforces it. The value is capped by a server-side per-task maximum, and requesting more raises `ApiError` (422) at submit time; that's documented on each `job_timeout_seconds` argument. A hardcoded client-side cap would duplicate a server policy that can change, and would go stale in either direction: too low and it blocks a job the server would have happily accepted, too high and it adds nothing.

So `_validate_job_timeout_seconds` only rejects what is invalid regardless of server policy (`<= 0`), which catches an obvious typo at the call site without spending a round-trip on it.

PR #867, opened 2026-08-26
fix: Snowflake finetune procedure fails to deploy on invalid default argument

gee-senbong:
## Problem

The Snowflake installer aborts while creating the finetune stored procedure, so `NIXTLA_FINETUNE` and the example datasets never get deployed:

```
000947 (42601): SQL compilation error:
invalid default argument expression CAST(1000 AS NUMBER(38,0))
```

## Cause

`create_finetune_sproc` registered the procedure with Snowpark's `@sproc` decorator. Snowpark renders Python defaults through `to_sql()`, whose `_IntegralType` branch unconditionally appends a cast, so `max_series: int = 1000` became `DEFAULT 1000 :: INT` — and Snowflake accepts only a bare literal there. (`params: Optional[dict] = None` renders as a plain `NULL`, which is why `MAX_SERIES` was the only argument that failed.)

Not a version regression: the rendering is identical across Snowpark 1.21.0–1.52.0, so pinning or upgrading Snowpark does not help.

## Fix

Emit the DDL directly and ship the handler as source instead of a cloudpickle blob, following the existing `TEMPLATE_SP` pattern:

```sql
"MAX_SERIES" NUMBER(38,0) DEFAULT 1000
```

Defaults now live only in the DDL — the handler signature declares none, since Snowflake resolves them at `CALL` time and always passes every argument. The body opens with a small import preamble so its annotations resolve in the module Snowflake execs it in.

Fixed as a consequence:

- Arguments are named `INPUT_DATA`/`PARAMS`/`MAX_SERIES` instead of `arg1`/`arg2`/`arg3`, so the named-argument `CALL` documented in `nbs/docs/deployment/3_snowflake.ipynb` works for the first time.
- Drops the implicit `cloudpickle>=3.1.1` requirement Snowpark injected.
- Public `CALL` signature is unchanged; 1-, 2- and 3-argument calls all still work.

## Behaviour change: `finetune_steps` default

The handler used to derive `finetune_steps` from the data when the caller omitted it:

```python
if "finetune_steps" not in params:
    params["finetune_steps"] = train_data["unique_id"].nunique()
```

This conflated gradient steps with series count. One series meant one step (and zero series meant zero steps — a silent no-op that still returned a model id), while a large table meant up to `MAX_SERIES` = 1000 steps, ~100× the library default, with cost scaling to match. It also made the same `CALL` non-reproducible as data grew.

Removed, so `NixtlaClient.finetune`'s own `finetune_steps=10` applies. Callers wanting a specific value use the documented route, `PARAMS => OBJECT_CONSTRUCT('finetune_steps', 100)`. Docs updated. No existing callers are affected — the procedure has never deployed successfully.

## Tests

`deploy_finetune=False` had been set in the test fixture since the Snowflake integration first landed, so this path had never run in CI. It is now enabled, `NIXTLA_FINETUNE` is checked by `verify_procedures_exist`, and `test_deploy_finetune_procedure` asserts via `GET_DDL` that the default is a bare literal rather than a cast.

### DIFF
--- a/nixtla/scripts/snowflake_install_nixtla.py
+++ b/nixtla/scripts/snowflake_install_nixtla.py
@@ -20,6 +20,7 @@ Options:
         - https://tsmp.nixtla.io (TimeGPT-2, supports all models)
 """
 
+import inspect
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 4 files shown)


nixtla/scripts/snowflake_install_nixtla.py
```diff
@@ -1339,10 +1340,66 @@ def create_stored_procedures(
...
+
+    from nixtla import NixtlaClient
+
+    token = _snowflake.get_generic_secret_string("nixtla_api_key")
+    base_url = _snowflake.get_generic_secret_string("nixtla_base_url")
+    client = NixtlaClient(api_key=token, base_url=base_url)
+
+    input_table = session.table(input_data)
+    ids = (
+        input_table.select("unique_id", F.hash("unique_id").alias("_od"))
```
JQGoh, line 1373: Q1: What is the purpose of `hash("unique_id")`? 

Q2: Am I right that let's say I have two different tables with the same set of unique series Ids but their order are different, this line of code can help us to suggest the same set of "ids" for finetuning? I think it is because of "order_by" operation.

JQGoh (approved):
Left 1 question to clarify on my understanding

nixtla/scripts/snowflake_install_nixtla.py
gee-senbong: Q1: The purpose of `hash("unique_id")` (combined with `order by`) is to produce a random sampling effect.

Q2: The `hash()` function in Snowflake is deterministic. Same input to the hash function produces the same output when we run it twice.

PR #883, opened 2026-09-11
[FEAT] Unify async job handling; add submit_simulate_job and submit_explain_job

gee-senbong:
`simulate()` and `explain()` shipped with their own async-job machinery — exponential backoff with jitter, client-level `async_job_wait_timeout`/`async_job_poll_interval`, and a `timeout_seconds` argument — while every other task used the shared submit-and-wait path. This PR puts them on that path.

## What changes

- New `submit_simulate_job()` and `submit_explain_job()` return a `Job` handle, the asynchronous counterparts to `simulate()` and `explain()`; `num_partitions` is not supported on them, use `simulate()` for that.
- The blocking calls are now built on the same submit-and-wait path, so both spellings send an identical request body.
- `timeout_seconds` becomes `job_timeout_seconds`, matching every `submit_*_job` method, and polling moves off the `NixtlaClient` constructor into per-call `poll_interval`/`poll_timeout` arguments, so `async_job_wait_timeout` and `async_job_poll_interval` are removed.
- Every task now polls the same way: a fixed interval, transient status-check failures bounded by the client's `max_retries`/`max_wait_time`, and `poll_timeout=None` for an unbounded wait.
- `AsyncJobError` carries the task name, derived from the route the job runs on, so a failed forecast reads `forecast job '<id>' failed: <error>`.

## Fixes

- `Job.__exit__` cancelled a job that `wait(cancel_on_timeout=False)` had deliberately left running, and sent a second cancel request whenever `wait()`'s own cancellation failed.
- `test_future_exog_alignment.py` still stubbed `_run_async_job`, which `simulate()` no longer calls, leaving two tests failing.

Nothing here touches a released API: `simulate`, `explain` and the removed constructor arguments all landed after `v0.9.0.dev1`. The `Unreleased` changelog section is rewritten to describe the 0.9.0 surface against `v0.8.0`, the last non-dev release.


elephaint (changes requested):
I didn't have anything, Claude made some remarks of which I kept the following. I mostly would re-think the fixed 15s polling time which for small panels can be annoying (maybe make polling time editable? Or change it based on the size of the payload based on simple heuristic?). 

- Bug: Job.__exit__ now skips cancellation based on exception type, not on which job raised it. With two handles open, a sibling's AsyncJobTimeoutError leaves the outer job running. Guard the early returns on exc_val.job_id == self.job_id.
- Bug: _wait_for_job_result_bytes with poll_timeout=None (now legal for execute_step) retries 5xx/connection errors with no cap. With poll_interval=0 it's a hot spin.
- Design trade-off to discuss: fixed 15s polling means a 2s explain() now returns in ~15s instead of ~3s. Also, forecast/finetune polls used to retry transient failures until the deadline; now ~75s of 502s ends the wait and, in partitioned fan-out, cancels the job and its siblings.
- Changelog: says submit_execute_step_job() returns a StepResult (it returns a Job), and overstates when the context manager cancels.
- Minor: SDK reference points at the new methods but has no entries for them. A few test gaps, notably the status-caching test passes regardless of whether caching works.

nixtla/async_job.py
```diff
@@ -283,13 +333,17 @@ def __exit__(self, exc_type, exc_val, exc_tb) -> None:
...
             return
-        if isinstance(exc_val, (AsyncJobCancelledError, AsyncJobTimeoutError)):
+        if isinstance(
+            exc_val, (AsyncJobCancelledError, AsyncJobTimeoutError)
+        ) and self._raised_by_this_job(exc_val):
             # `wait()` already settled both: a cancelled job is terminal, and a
             # timed-out one was either cancelled there or deliberately left
-            # running by `cancel_on_timeout=False`.
+            # running by `cancel_on_timeout=False`. A sibling's timeout settles
+            # nothing here, so it falls through to the cleanup below.
```
gee-senbong, line 342: @elephaint This attempts to address the Job exit related comments.

gee-senbong:
> I didn't have anything, Claude made some remarks of which I kept the following. I mostly would re-think the fixed 15s polling time which for small panels can be annoying (maybe make polling time editable? Or change it based on the size of the payload based on simple heuristic?).
> 
> * Bug: Job.**exit** now skips cancellation based on exception type, not on which job raised it. With two handles open, a sibling's AsyncJobTimeoutError leaves the outer job running. Guard the early returns on exc_val.job_id == self.job_id.
> * Bug: _wait_for_job_result_bytes with poll_timeout=None (now legal for execute_step) retries 5xx/connection errors with no cap. With poll_interval=0 it's a hot spin.
> * Design trade-off to discuss: fixed 15s polling means a 2s explain() now returns in ~15s instead of ~3s. Also, forecast/finetune polls used to retry transient failures until the deadline; now ~75s of 502s ends the wait and, in partitioned fan-out, cancels the job and its siblings.
> * Changelog: says submit_execute_step_job() returns a StepResult (it returns a Job), and overstates when the context manager cancels.
> * Minor: SDK reference points at the new methods but has no entries for them. A few test gaps, notably the status-caching test passes regardless of whether caching works.


The following has been updated:
* Job exit cancellation logics has been revised.
* The fixed internal polling logic has been updated to exponential backoff
* Change log is updated to reflect current state

elephaint (approved):
Nice!

PR #887, opened 2026-09-15
[REFACTOR] Move async jobs into a client.jobs namespace and a nixtla/jobs package

gee-senbong:
## What

The seven `submit_*_job()` methods on `NixtlaClient` become a `client.jobs` namespace, and every job-related module moves out of `nixtla_client.py` into `nixtla/jobs/`.

```python
client.jobs.forecast(...)            # was submit_forecast_job
client.jobs.cross_validation(...)    # was submit_cross_validation_job
client.jobs.finetune(...)            # was submit_finetune_job
client.jobs.detect_anomalies(...)    # was submit_anomaly_detection_job
client.jobs.simulate(...)            # was submit_simulate_job
client.jobs.explain(...)             # was submit_explain_job
client.jobs.execute_step(...)        # was submit_execute_step_job
```

Each name now mirrors the blocking method it matches, so it can be derived rather than memorised — note that `jobs.detect_anomalies()` mirrors `detect_anomalies_online()`, i.e. it runs *online* detection.

| module | holds |
|---|---|
| `jobs/_job.py` | the `Job` handle, `JobStatus` and the errors (was `async_job.py`) |
| `jobs/_namespace.py` | `Jobs` — the seven `client.jobs` methods |
| `jobs/_transport.py` | submit / poll / cancel / result fetch |
| `_http.py` | `ApiError`, retriable-error classification, the logger |

`_http.py` is split out so the transport can classify errors without importing the client back; it still logs on `nixtla.nixtla_client`, so the channel callers configure is unchanged. `steps.py` becomes `_steps.py` — it is the `execute_step` codec, reached only through `jobs.execute_step()`, so it belongs with the other underscore-prefixed internals, while `StepResult` and `ref` stay exported from the package root.

## Testing

`ruff` and `mypy` are clean. The full non-distributed suite reproduces the `main` baseline exactly: the same 1 failure, 13 skipped and 21 errors in both runs, every one of them caused by unset `NIXTLA_*_CUSTOM` environment variables rather than by this change, with 751 passing here against 746 on `main` — the five extra being new tests covering the namespace. Import smoke tests cover `import nixtla`, `import nixtla.jobs`, `from nixtla.jobs import _transport`, `from nixtla.nixtla_client import ...` and `from nixtla._steps import ...` in fresh interpreters, since the package now has an import cycle whose resolution depends on module order.


nixtla/__init__.py
```diff
@@ -1,21 +1,21 @@
 from importlib.metadata import version
 
-from .async_job import (
-    AsyncJobCancelledError,
-    AsyncJobError,
-    AsyncJobTimeoutError,
+from .jobs import (
```
nasaul, line 3: The import cycle this introduces is load-bearing and nothing in CI guards it. Two orderings have to hold:

- here, `.jobs` before `.nixtla_client`
- in `jobs/_namespace.py`, `from . import _transport` before `from ..nixtla_client import ...`

Swap either and the package stops importing entirely. The repo's ruff config selects only `F` (pyflakes) with no isort rules, so nothing reorders these today — but nothing catches it either, and the failure is total rather than subtle.

You mention running import smoke tests; worth committing them so the constraint is enforced rather than remembered:

```python
@pytest.mark.parametrize(
    "stmt",
    ["import nixtla", "import nixtla.nixtla_client", "import nixtla.jobs._namespace"],
)
def test_imports_in_a_fresh_interpreter(stmt):
    subprocess.run([sys.executable, "-c", stmt], check=True)
```

nixtla/jobs/_job.py
```diff
@@ -204,17 +215,39 @@ def status(self) -> JobStatus:
...
-                http_client, self._endpoint, self.job_id
+            job_data = get_job_data(
+                self._client, http_client, self._endpoint, self.job_id
             )
         status = JobStatus(job_data.get("status"))
         if status.is_terminal:
             self._status = status
         return status
 
+    def refresh(self) -> "Job":
```
nasaul: `Job.refresh()` is new here and has no callers inside the package — `status` goes to `_refresh_status()` directly, so `refresh()` is purely public surface. Its docstring then concedes that the chaining it exists for (`job.refresh().status`) costs two requests on a running job, where `job.status` alone costs one and does the same refresh.

Suggest cutting it from this PR rather than documenting it in `sdk_reference.mdx` (where it is currently missing — the `## Job` section lists only `wait` and `cancel`). If you want the explicit-refresh idiom, the version that pays for itself is the inverse — `status` as a cached attribute, `refresh()` as the only call that hits the network — but that changes `status` semantics and deserves its own PR.

nixtla/nixtla_client.py
```diff
@@ -1723,6 +1466,20 @@ def __init__(
...
 
+    @functools.cached_property
+    def jobs(self) -> "Jobs":
+        """Namespace for work that runs server-side: `client.jobs.forecast(...)`.
+
+        Each method there mirrors the blocking method of the same name but
+        returns a `Job` handle instead of a result. See `Jobs`.
+        """
+        # Deferred: `jobs._namespace` imports this module at its top. Building
+        # it lazily also keeps the back-reference out of `__dict__` for clients
```
nasaul: The second sentence of this comment doesn't hold up. `NixtlaClient` isn't picklable by stdlib `pickle` at all — `AttributeError: Can't pickle local object 'retry.<locals>.wrap'` — and that's already true on `main`, before this PR. The distributed backends use `cloudpickle`, which handles the reference cycle fine: I round-tripped a client with `.jobs` materialised and the payload grew 60 bytes, with `jobs._client` correctly rebound to the copy.

The first sentence is the real and sufficient reason. Suggest just dropping the rest:

```python
# Deferred: `jobs._namespace` imports this module at its top.
```

nasaul (approved):
Approving — this is a faithful move. I diffed every relocated function against `main`: all seven job methods and the eleven transport helpers are identical modulo `self` → `nixtla_client`/`self._client` and the `AsyncJob*Error` → `Job*Error` renames. The four-layer split with `_http.py` as the dependency-free leaf is the right shape, and dropping the deprecation shims is correct given `nixtla/steps.py` and the `submit_*_job` methods only ever shipped in `0.9.0.dev0/dev1`.

Verified locally: 328 job/step/simulate tests pass, `ruff` clean, all six import entry points work in fresh interpreters, `client.jobs` caches correctly, and `cloudpickle` round-trips a client with `.jobs` materialised.

Three comments inline, none blocking. Only the first is worth doing before merge.

PR #888, opened 2026-09-15
[FEAT] Add `client.jobs.list()` and `client.jobs.retrieve()` to recover async jobs

gee-senbong:
## What

`client.jobs.list()` lists your team's jobs newest-first, with `status`, `task` and `limit` filters, paging until `next_page_token` is null rather than until a page looks short. `client.jobs.retrieve(job_id)` turns an id back into a working `Job` — `status`, `wait()`, `cancel()` and the context manager all behave as on a submitted one — so a job can outlive the process that submitted it. The task is read from the id's own prefix, so `retrieve()` costs no round trip.

## Why `retrieve()` returns a raw dict

A retrieved `wait()` returns the server's raw result dict, not a parsed dataframe. Every `parse_result` closes over request-side state, and the series ids and column names never cross the wire, so they cannot come back with the result. Values run in sorted series-id order, so a caller still holding the input frame can re-label them. `execute_step` is the exception — its result is self-describing, so `wait()` still returns a `StepResult`.

## Tests

19 mocked tests cover paging (including a short or empty page that still carries a token), prefix resolution for all seven tasks, and the raw-result contract. `test_jobs_integration.py` adds 12 live tests over `forecast`, `cross_validation`, `detect_anomalies`, `finetune` and `execute_step`: lifecycle, cancellation, poll-timeout resumability, context-manager cleanup, and the 422 for a `job_timeout_seconds` above the server's 300s cap. These are the first live tests for `client.jobs`. Full run: 358 mocked passed, 12 live passed.

The listing is backed by Temporal's visibility index and trails submit by about a second, so the live tests poll until a job appears rather than asserting once.


nasaul (commented):
One thing to fix before merge, three small ones, and ~50 lines that can come out.

**`limit` stops bounding the walk once `task` is set** — `_namespace.py:L1011`

`page_size` is suppressed when `task is not None`, and the early return only fires after `limit` *matching* rows accumulate. So `jobs.list(task="explain", limit=1)` on a team whose recent jobs are all forecasts pages through the entire retention window at the server-default page size, one request per page.

The integration helper depends on the opposite: `LISTING_SCAN_LIMIT = 500` is commented "Caps the walk", but `_list_until_present(..., task="forecast")` passes it *with* a task filter, so nothing caps it — on a busy team that's potentially hundreds of requests per poll iteration, every second, for up to `LISTING_LAG_TIMEOUT`. Bounding rows *fetched* (or pages) independently of the client-side filter fixes both, and deletes the nested `page_size` conditional as a side effect.

**Smaller things**

- `_transport.py:L115` — `list_jobs` calls `_get_request` bare, while `list_models` (`nixtla_client.py:2008`) and `_finetuned_models` (`nixtla_client.py:2177`) both go through `self._retry_strategy(self._get_request)`. One transient 502 on page 7 of a multi-page walk raises and discards every row already collected, with no resume token for the caller. One-line wrap.
- `_namespace.py:L1031` — `JobStatus(row["status"])` raises a bare `ValueError` and `row["created_at"]` a `KeyError`, so a single unmodelled status (a queued or timed-out state the enum doesn't have) fails the whole listing, including the caller's unrelated rows. `poll_job` is careful here — catches `ValueError`, re-raises as a `JobError` naming the job — and `task_name` is already tolerated via `.get`.
- `_namespace.py:L1001` — `task` accepts a plain `str`, so `status` reads as though it does too. `status="succeeded"` iterates characters and reports `ValueError: 's' is not a valid JobStatus`, which points at nothing the caller wrote. Normalising a bare str/enum to a single-element list is one line.

**Duplication worth collapsing (~50 lines)**

- `_transport.py:L546-593` — `wrap_retrieved_job`'s two `get_result` closures are verbatim copies of the ones at L511 and L671: same `JobError` guard, same `wait_for_job_result_bytes` + `_build_step_result` body, same `Job(...)` construction. One `_wrap_job(nixtla_client, job_id, endpoint, task, parse_result)` covers all three; the callers then differ only by whether they submit first, with `retrieve()` passing identity.
- `_transport.py:L79` — `_TASK_ENDPOINTS` is a dict built to produce `f"v2/{task}"`, and it's the inverse of `_task_name()` twenty lines above. Inline the f-string, validate `task` against `_JOB_ID_PREFIXES.values()`, keep the `anomaly_detection` comment.
- `_transport.py:L83` — `_BINARY_TASKS = frozenset({"execute_step"})` has one member and one membership test.

gee-senbong:
All six addressed.

- **`limit` stops bounding the walk once `task` is set** — `limit` now caps rows *fetched*, so `page_size` is capped whether or not `task` is set and shrinks to the unfetched remainder (`min(limit - fetched, _MAX_PAGE_SIZE)`). The nested conditional is gone. Documented as a semantic change: with `task` set, fewer than `limit` rows can come back. An empty page charges 1 against the budget too — the server serves empty pages with a live `next_page_token`, and without that they cost a request but no budget, which reopened the same unbounded walk through the server-side `status` filter. `LISTING_SCAN_LIMIT = 500` in the integration helper is now a real bound (≤3 requests per poll), and its comment says so.
- **`list_jobs` bare `_get_request`** — wrapped in `_retry_strategy`, like `list_models`, so a transient failure on page 7 no longer discards the pages already collected. One note: the review says `_finetuned_models` also wraps it — it doesn't, it calls `_get_request` bare. That predates this PR so I left it; happy to fix it separately if you want.
- **Unmodelled status / missing `created_at`** — rows this client cannot read are skipped with a warning naming the row rather than failing the page; `JobSummary.created_at` is now `Optional[str]`. Called out in the docstring and the changelog, since for a recovery API "absent here" now has two meanings.
- **Bare `status` string** — normalised to a one-element list. `JobStatus` subclasses `str`, so one `isinstance` covers both forms.
- **`wrap_retrieved_job` duplication** — one `_wrap_job(nixtla_client, job_id, endpoint, task, parse_result)`; all three callers route through it, `retrieve()` and the binary path passing identity.
- **`_TASK_ENDPOINTS` / `_BINARY_TASKS`** — both deleted; endpoint inlined as `f"v2/{task}"`, `jobs.list()` validates against `_JOB_ID_PREFIXES.values()`, and the `anomaly_detection` note moved onto that map. `_transport.py` is net -22 lines.

Offline suite is 292 passing (was 280).

JQGoh (approved):

PR #892, opened 2026-09-16
[REFACTOR] Extract _types.py and _audit.py; run live-API tests on one CI cell

gee-senbong:
## What's here

**`nixtla/_types.py` (new, 119 lines)** — type aliases, `validate_extra_params`, `FinetunedModel` and the endpoint constants. Imports nothing from the package, so it sits at the bottom of the import graph. This shrinks the `jobs._namespace` → `nixtla_client` runtime edge to two preprocessing helpers; once `_preprocessing.py` is extracted, the import cycle that `nixtla/__init__.py`'s statement order currently tiptoes around should disappear entirely.

**`nixtla/_audit.py` (new, 350 lines)** — the D001/D002/F001/V001/V002 checks and the `audit_data`/`clean_data` implementations. Pure local pandas, no API access. `NixtlaClient.audit_data` and `.clean_data` keep their signatures and docstrings and forward to it, so the public API is unchanged.

Both are pure moves. The only edits to moved code were mechanical: a dedent, dropping `@staticmethod`/`self`, and one `self.audit_data(` → `audit_data(`. `_audit.py` is AST-identical to #886's version.

**Snowflake demo notebook** moved from `nixtla/snowflake/` to `nbs/docs/deployment/4_snowflake_anomaly_demo.ipynb`. Because the build backend uses `module-root = "."`, it had been shipping inside every installed wheel; `uv build` now produces a wheel with no `.ipynb`. It's excluded from the nbconvert job, since `get_active_session()` only resolves inside a Snowflake worksheet.

**Live-API tests now run on one matrix cell.** A `pytest_collection_modifyitems` hook marks any test taking a live-client fixture as `integration`, so the marker follows the fixture that actually reaches the API rather than being hand-applied to ~170 tests. ubuntu-latest/3.11 still runs everything; the other five cells run `-m \"not integration and not distributed_run and not snowflake\"`. Collection goes 875 → 690 there.

## Worth a reviewer's attention

Verifying that the reduced selection is *genuinely* offline — by pointing the base URL at a closed port — turned up **11 tests that were hitting the live API on all six runners with no marker at all**: five in `test_client.py` that build their own `NixtlaClient()`, and all of `test_retry.py`, whose `check_retry_behavior` helper mocks `_make_request` but not the `_get_request` that `forecast()` uses for model-metadata lookup. Those are now marked explicitly.

Coverage of the offline selection alone is 85%, so the five non-primary cells still clear `--cov-fail-under=80`.

## Follow-ups (deliberately not fixed here — all would have broken the pure-move mandate)

- **Real bug:** `clean_data`'s re-audit drops `target_col` (`nixtla/_audit.py:346`), so `clean_data(..., target_col=\"value\")` cleans `value` then re-audits `y`. Inherited verbatim.
- `DataFrameGroupBy.apply` needs `include_groups=False` before the `pandas<3.0.0` pin lifts.
- Stubbing `_get_model_params` in `check_retry_behavior` would let `test_retry.py` drop its `integration` marker — retry/backoff is pure client logic worth running on every cell.
- `_audit.py` imports `._http` only to borrow `logger`; moving `logger` somewhere neutral would tidy the layering.
- `addopts` still carries a stale `--cov=python/statsforecast`, which is why local single-test runs need `--no-cov`.

## Testing

`836 passed, 13 skipped, 1 xpassed, 0 failed`, coverage 90.33% (`-m "not distributed_run and not snowflake"`, macOS/3.11). ruff clean; mypy at the pinned pre-commit revision reports no issues.


.github/workflows/pytest.yml
```diff
@@ -72,11 +72,15 @@ jobs:
...
+      # automatically -- see `pytest_collection_modifyitems` in the conftest.
+      - name: Run offline tests (not Linux Python 3.11)
         if: runner.os != 'Linux' || matrix.python-version != '3.11'
         env:
           NIXTLA_API_KEY: ${{ secrets.NIXTLA_DEV_API_KEY }}
           NIXTLA_BASE_URL: ${{ secrets.NIXTLA_DEV_BASE_URL }}
           NIXTLA_API_KEY_CUSTOM: ${{ secrets.NIXTLA_API_KEY_CUSTOM }}
           NIXTLA_BASE_URL_CUSTOM: ${{ secrets.NIXTLA_BASE_URL_CUSTOM }}
-        run: uv run pytest --cov=nixtla -m "not distributed_run and not snowflake" --reruns 3 --reruns-delay 10 --only-rerun ConnectError --only-rerun "Too Many Requests" --only-rerun "Internal server error" nixtla_tests
+        run: uv run pytest --cov=nixtla -m "not integration and not distributed_run and not snowflake" --reruns 3 --reruns-delay 10 --only-rerun ConnectError --only-rerun "Too Many Requests" --only-rerun "Internal server error" nixtla_tests
```
gee-senbong, line 86: Update this part so that the integration tests are not run for every OS & python version

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 
 - **Breaking:** `detect_anomalies_online()` takes its two new arguments mid-signature, matching the parameter order `forecast()` and `cross_validation()` use. Callers passing arguments **positionally** past `finetune_loss` must switch to keywords. Its request body now always carries `finetuned_model_id`, which was previously omitted when unset.
 - A failed partition now stops queued `num_partitions` work instead of letting the remaining requests run to completion.
+- The built wheel no longer ships a notebook. The Snowflake anomaly-detection demo moved out of the package to `nbs/docs/deployment/4_snowflake_anomaly_demo.ipynb`; because the build backend is configured with `module-root = "."`, it had been installed alongside the library by every `pip install nixtla`.
```
JQGoh, line 19: @ngupta23 FYI

JQGoh (approved):

CHANGELOG.md
ngupta23: Makes sense. @gee-senbong and I discussed it before the change was made. Thanks!

PR #896, opened 2026-09-17
[CHORE] Warn about the 1.0 anomaly-detection rename

gee-senbong:
In 1.0 the batch `detect_anomalies()` is removed and the name is reused for online detection, so the same call keeps working but means something different. Add `FutureWarning`s on every entry point a user can reach:

- `detect_anomalies()` -- removal notice; no drop-in replacement
- `detect_anomalies_online()` -- rename notice; behavior unchanged
- `nixtla.snowflake.detect_anomalies()` -- switches to online detection, needs `h`/`detection_size`, requires a re-deploy

`jobs.detect_anomalies()` stays silent: it already runs online detection under the 1.0 name, so nothing using it has to change. Its docstring now says so instead of contrasting itself with `detect_anomalies_online()`.

`FutureWarning` rather than `DeprecationWarning` -- the latter is filtered out by default outside `__main__`, which is exactly where these users are (notebooks, pipeline scripts). `stacklevel=2` throughout, and each warning is the first statement in its method so it fires in the caller's process rather than only in a distributed worker.

No warning inside `nixtla_detect_anomalies_batch`; it runs in the warehouse where nothing reads stderr. The installer prints the notice at deploy time instead, which is when the operator can act on it.

Keeping `h` and `detection_size` required with no defaults is what makes stale batch calls raise `TypeError` at 1.0 rather than silently return a different detection window.


nixtla/snowflake/anomaly.py
```diff
@@ -178,7 +188,24 @@ def detect_anomalies(
...
+        ``NixtlaClient.detect_anomalies_online()``). ``h`` and
+        ``detection_size`` become required ``params``, and the result covers a
+        trailing detection window rather than the whole history.
+
     """
+    warnings.warn(
+        "nixtla.snowflake.detect_anomalies() and the NIXTLA_DETECT_ANOMALIES "
+        "procedure currently perform historical (batch) anomaly detection. In "
+        "nixtla 1.0 they switch to online detection, matching "
+        "NixtlaClient.detect_anomalies_online(): `h` and `detection_size` "
```
JQGoh: may be `"NixtlaClient.detect_anomalies_online() of <=v0.9.0`, mentioning the version explicitly?

JQGoh (approved):
LGTM, thanks for introducing the deprecation warning at this stage. I left a minor suggestion

nixtla/snowflake/anomaly.py
gee-senbong: Good catch — pinned it. Replaced every "today's `detect_anomalies_online()`" with "`detect_anomalies_online()` as of v0.9.0", in the Snowflake warning, both Snowflake docstring notes, and the `NixtlaClient.detect_anomalies()` warning. e12eeda

PR #897, opened 2026-09-18
[DOCS] Add an asynchronous jobs guide

gee-senbong:
## What

Adds a narrative guide for async jobs at `/jobs/asynchronous_jobs`, under a new
**ASYNCHRONOUS JOBS** navigation group.

`reference/sdk_reference.mdx` already documents every `client.jobs` method,
`Job`, `Job.wait()`, `JobStatus` and the error types. What was missing was a page
that teaches the pattern — before this, async jobs appeared in the nav-visible
docs only as a `<Note>` on the simulation page.

## Where it sits

The group goes after **ANOMALY DETECTION** rather than inside **FORECASTING**,
because jobs are cross-cutting: they cover forecasting, cross-validation,
anomaly detection, fine-tuning, simulation and explanation.

## What the page covers

- What a `Job` handle is and why you would want one
- Submitting, reading `status`, and `wait()`, as numbered steps on AirPassengers
- The two independent limits — `poll_interval` / `poll_timeout` /
  `cancel_on_timeout` on the client side, `job_timeout_seconds` on the server
- `cancel()` and the context manager
- `JobError`, `JobCancelledError`, `JobTimeoutError`
- Fanning out over several submissions
- A table mapping each `jobs.*` method to the blocking one it mirrors

Per-argument tables are deliberately left to the SDK reference; the guide links
across rather than duplicating them.

## Verification

Every snippet was run verbatim against the live API — the job ids, statuses, log
lines and result tables on the page are its real output. `docs.json` parses, all
67 nav pages resolve to files, and all 7 cross-links in the new page hit real
nav entries.

Two things surfaced only by running the snippets, and are reflected in the page:

- `job_timeout_seconds` above the per-task server maximum raises `ApiError`
  (422) **at submission**, not as a job failure at `wait()`. The page's example
  uses a value under the limit and shows the 422 as the failure mode.
- The status immediately after submit is `pending` as often as `running`, so the
  page shows `pending` and its comparison examples no longer assert a value that
  depends on timing.

## Not included

`jobs.list()` and `jobs.retrieve()` (no longer in the SDK) and
`jobs.execute_step()` (on the namespace, but with no reference entry or nav
presence yet).


timegpt-docs/jobs/asynchronous_jobs.mdx
```diff
@@ -0,0 +1,309 @@
...
+
+## Available Job Methods
+
+| Method | Mirrors | `wait()` returns |
+|:---|:---|:---|
+| `jobs.forecast()` | [`forecast()`](/forecasting/timegpt_quickstart) | Forecasts as a DataFrame |
+| `jobs.cross_validation()` | [`cross_validation()`](/forecasting/evaluation/cross_validation) | Cross-validation results as a DataFrame |
+| `jobs.detect_anomalies()` | [`detect_anomalies_online()`](/anomaly_detection/real-time/introduction) | Anomalies as a DataFrame |
+| `jobs.finetune()` | [`finetune()`](/forecasting/fine-tuning/steps) | The fine-tuned model id, as a string |
+| `jobs.simulate()` | [`simulate()`](/forecasting/probabilistic/simulation) | Sample paths as a DataFrame |
```
JQGoh, line 299: maybe "Simulated sample paths", with the first word simulated can clarify the intension better?

timegpt-docs/jobs/asynchronous_jobs.mdx
```diff
@@ -0,0 +1,309 @@
...
+| 1958-05-01 | 1957-12-01 | 363 | 385.46 |
+
+Both jobs were already running by the time the first `wait()` was called, so
+the second result costs only whatever time it still needed.
+
+## Available Job Methods
+
+| Method | Mirrors | `wait()` returns |
+|:---|:---|:---|
+| `jobs.forecast()` | [`forecast()`](/forecasting/timegpt_quickstart) | Forecasts as a DataFrame |
```
JQGoh, line 295: can I double check that /forecasting/timegpt_quickstart etc., are these links available in the public already?

Just want to make sure it is not a broken link

JQGoh (approved):

timegpt-docs/jobs/asynchronous_jobs.mdx
gee-senbong: Tested locally and it works properly.

timegpt-docs/jobs/asynchronous_jobs.mdx
JQGoh: Understood. Sample paths in this case, the word "sample" itself is a verb than noun.

timegpt-docs/jobs/asynchronous_jobs.mdx
JQGoh: Validated that these are valid links

PR #898, opened 2026-09-18
[REFACTOR] Split _preprocessing.py and _payloads.py out of nixtla_client.py

gee-senbong:
## What's here

The last two pieces of the split that produced `_types.py`, `_audit.py`, `_http.py` and the `jobs/` package. `nixtla_client.py` goes **5052 → 3045 lines**, with no public API change: `nixtla.__all__` is unchanged and every name that moved is underscore-prefixed.

- **`nixtla/_preprocessing.py`** (new) — the 42 dataframe helpers: validation, frequency and timezone handling, exogenous features, partitioning, response parsing. It imports only `._http` and `._types`.
- **`nixtla/_payloads.py`** (new) — the six `prepare_*` builders, now free functions taking the client, so the blocking methods and `client.jobs` share one builder per task and cannot drift.
- **The `jobs/_namespace.py` → `nixtla_client` runtime import edge is gone** — the two helpers it reached for were preprocessing all along — which closes the cycle that `nixtla/__init__.py`'s statement order was tiptoeing around. `nixtla_tests/test_imports.py` now asserts the full layering: `_types`/`_http` → `_preprocessing` → `_payloads` → `nixtla_client`.

## Worth a reviewer's attention

Both extractions are pure moves, verified mechanically rather than by eye: every definition was re-extracted from the base commit by AST, normalized, `ruff format`ed and diffed — 42/42 and 6/6 names identical, and of ~90 surviving definitions in `nixtla_client.py` exactly six changed, each only at its call site.

Batch `detect_anomalies()` is the one task without a builder: it never had a `_prepare_*` method to move and is being removed in 1.0, so its payload is still assembled inline in `nixtla_client.py`.

`test_start_datetime_payload.py` and `test_categorical_features.py` reach `_partition_series` and `_extract_categorical_exog` through the `nixtla_client` module namespace; both names stay bound there and are not re-imported by `_payloads`, so those monkeypatches still intercept every call.

## Testing

`674 passed, 1 skipped, 174 deselected` (`-m "not distributed_run and not integration"`, macOS/3.12) against a `665 passed, 1 skipped` baseline — the +9 are the new layering tests. `ruff check nixtla/` is clean and mypy reports no issues.


ngupta23 (approved):
Reviewed. Found several reuse/coupling issues in the new `_payloads.py`/`_preprocessing.py` (memory-pinning closures, duplicated future-exog logic, divergent hist_exog exclusion forms, duplicated duplicate-key/positive-int validation, hand-rolled group boundaries). Verified all of these are pre-existing in the old monolithic `nixtla_client.py` and were moved verbatim by this refactor rather than introduced by it, so not blocking. Filed as follow-up tech debt in #899.

Approving as-is.

